PR #1, opened 2024-04-01
Add object models with PyPOM

ben-c-at-moz:
* Use PyPOM to update object models for About pages and Browser navigation
* Parametrize the search mode test
* Create a naming hierarchy and imports for object models
* Linting and formatting
* Add "pynput" mark for tests that use it--we'll have our work cut out using it in headless mode
* Update some of the hook installer scripts (please ignore for now)


tests/credential/conftest.py
```diff
@@ -1,12 +1,9 @@
...
+from faker import Faker
+from faker.providers import internet, misc
 from selenium.webdriver import Firefox
-from selenium.common.exceptions import WebDriverException
 
-from modules.util import BrowserActions
-from modules.shadow_dom import AboutLogins
-from faker import Faker
-from faker.providers import internet
-from faker.providers import misc
```
pdehaan, line 9: :+1: Good catch. Did you just 👀 this, or did some linter tooling catch it?

tests/credential/test_about_logins_search.py
```diff
@@ -1,15 +1,15 @@
+from typing import List, Tuple
+
 from selenium.webdriver import Firefox
+
+from modules.page_object import AboutLogins
 from modules.util import BrowserActions
-from modules.shadow_dom import AboutLogins
-from typing import Tuple
-from typing import List
```
pdehaan, line 5: 😍

conftest.py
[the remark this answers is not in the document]
ben-c-at-moz: That's probably a good idea. I'll look at what other people have done.

tests/credential/conftest.py
ben-c-at-moz: that's ruff!

conftest.py
```diff
@@ -23,12 +23,24 @@ def pytest_addoption(parser):
         help="Run in headless mode: --run-headless=True",
     )
 
+    parser.addoption(
+        "--implicit-timeout",
+        action="store",
+        default=10,
+        help="Timeout for implicit waits, set 0 for no wait (default 10)",
```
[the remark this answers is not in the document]
ben-c-at-moz: So, implicit wait is just "hey how long do we wait when doing things like `find_element`"? I set the default to 10, but I messed up later in this file, which you also caught.

conftest.py
```diff
@@ -79,6 +96,7 @@ def driver(fx_executable: str, opt_headless: bool, set_prefs: List[Tuple]):
     for opt, value in set_prefs:
         options.set_preference(opt, value)
     s = webdriver.Firefox(options=options)
+    s.implicitly_wait(10)
```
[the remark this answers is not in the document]
ben-c-at-moz: this should have been the value returned by the fixture above. Will fix.

.gitignore
```diff
@@ -57,6 +58,9 @@ coverage.xml
 .pytest_cache/
 cover/
 
+# pytest-html
+assets/
```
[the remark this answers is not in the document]
ben-c-at-moz: Good catch, fixing...

conftest.py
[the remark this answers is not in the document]
ben-c-at-moz: I'm seeing most newer projects moving away from the per-file header. Looks like most of them are content to have the MPL in their root.

conftest.py
```diff
@@ -79,6 +96,7 @@ def driver(fx_executable: str, opt_headless: bool, set_prefs: List[Tuple]):
     for opt, value in set_prefs:
         options.set_preference(opt, value)
     s = webdriver.Firefox(options=options)
+    s.implicitly_wait(opt_implicit_timeout)
```
jrbenny35, line 99: `s` doesn't make much sense here. Could you name this something or maybe just use `_`?

modules/browser_object_navigation.py
```diff
@@ -0,0 +1,109 @@
...
+    # Search Mode One-Offs
+    _search_one_off_settings_button = (By.ID, "urlbar-anon-search-settings")
+
+    def ensure_chrome_context(self):
+        if self._xul_source_snippet not in self.driver.page_source:
+            self.driver.set_context(self.driver.CONTEXT_CHROME)
+
+    def resume_content_context(self):
+        if self._xul_source_snippet in self.driver.page_source:
+            self.driver.set_context(self.driver.CONTEXT_CONTENT)
```
jrbenny35, line 46: I am interested in why you chose to do the chrome context setting this way. Couldn't you just use the generator selenium provides for this? That way you couldn't have to manage switching back to the `content` context

modules/browser_object_navigation.py
```diff
@@ -0,0 +1,109 @@
...
+        return self
+
+    def set_search_mode_via_awesome_bar(self, mode: str) -> WebElement:
+        if mode in self.BROWSER_MODES:
+            abbr = self.BROWSER_MODES[mode]
+        else:
+            abbr = mode.lower()[:2]
+        self.type_in_awesome_bar(abbr)
+        self.wait.until(EC.visibility_of_element_located(self._tab_to_search_text_span))
+        self.awesome_bar.send_keys(Keys.TAB)
```
jrbenny35, line 66: Why are you not using `type_in_awesome_bar`?

modules/browser_object_navigation.py
```diff
@@ -0,0 +1,109 @@
...
+    fx_refresh_menu_get_help_item = (
+        By.CSS_SELECTOR,
+        'menuitem[data-l10n-id="urlbar-result-menu-tip-get-help"]',
+    )
+
+    # Search with...
+    search_engine_suggestion_row = (
+        By.CSS_SELECTOR,
+        'div[class="urlbarView-row"][type="search_engine"]',
+    )
```
jrbenny35, line 109: Why are these at the bottom of this file?

modules/browser_object_navigation.py
```diff
@@ -0,0 +1,109 @@
+from pypom import Page
+from selenium.webdriver.common.by import By
+from selenium.webdriver.common.keys import Keys
+from selenium.webdriver.remote.webelement import WebElement
+from selenium.webdriver.support import expected_conditions as EC
+
+
+class Navigation(Page):
```
jrbenny35, line 8: Is this class really a page, or more of a Region?

modules/page_object_about_prefs.py
```diff
@@ -0,0 +1,73 @@
...
+    """Page Object Model for about:preferences"""
+
+    URL_TEMPLATE = "about:preferences#{category}"
+
+    def __init__(self, driver, **kwargs):
+        super().__init__(driver, timeout=10, **kwargs)
+        self.utils = PomUtils(self.driver)
+
+    class Dropdown(Region):
+        _active_dropdown_item = (By.CSS_SELECTOR, "menuitem[_moz-menuactive='true']")
```
jrbenny35, line 24: I'd suggest having a property reference this region so your function call would be like `about.dropdown.function()`. Pretty much what you have on L55-57

tests/address_bar_and_search/test_search_modes_for_sites.py
```diff
@@ -1,26 +1,21 @@
...
 from selenium.webdriver.support.wait import WebDriverWait
-from modules.page_object import Navigation
 
+from modules.browser_object import Navigation
 
-def test_search_modes_for_sites(driver: Firefox, search_modes: dict):
+sites = ["Google", "Amazon", "Bing", "DuckDuckGo", "eBay"]
+
+
+@pytest.mark.parametrize("site", sites)
```
jrbenny35, line 12: 🥳

tests/address_bar_and_search/test_search_modes_for_sites.py
```diff
@@ -1,26 +1,21 @@
...
-            awesome_bar.send_keys("soccer" + Keys.RETURN)
-            with driver.context(driver.CONTEXT_CONTENT):
-                wait.until(EC.url_contains(site.lower()))
-            awesome_bar.clear()
+    nav = Navigation(driver).open()
+    nav.set_search_mode_via_awesome_bar(site)
+    nav.type_in_awesome_bar("soccer" + Keys.RETURN)
+    with driver.context(driver.CONTEXT_CONTENT):
+        nav.wait.until(EC.url_contains(site.lower()))
+    nav.awesome_bar.clear()
```
jrbenny35, line 21: Look at how tidy this is, so nice!

tests/credential/conftest.py
```diff
@@ -39,21 +36,19 @@ def driver_and_saved_usernames(driver: Firefox, faker: Faker, origins):
...
-            about_logins.add_login_button()
-        except WebDriverException:
-            about_logins.login_item_save_changes_button().click()
+        about_logins.create_new_login(
+            {
+                "origin": origin,
+                "username": username,
+                "password": password,
+            }
+        )
```
jrbenny35, line 51: Not needed but `pytest-variables`

modules/browser_object_navigation.py
jrbenny35: Also, you didn't create an `__init__` here, why is that?

modules/page_object_about_prefs.py
```diff
@@ -0,0 +1,73 @@
...
+
+        @property
+        def loaded(self):
+            return self.root if EC.element_to_be_clickable(self.root) else False
+
+        def select_option(self, option_name):
+            if not self.dropmarker.get_attribute("open") == "true":
+                self.root.click()
+            matching_menuitems = [
+                e
```
jrbenny35, line 42: I like using `el` for short length format of `element`

jrbenny35 (changes requested):
Added some thoughts. I also noticed you don't initialize all the pages, could you explain why?

jrbenny35:
> Add "pynput" mark for tests that use it--we'll have our work cut out using it in headless mode

Have you had problems with headless mode?

ben-c-at-moz:
> > Add "pynput" mark for tests that use it--we'll have our work cut out using it in headless mode
> 
> Have you had problems with headless mode?

Yep, so far I'm not able to use any tests that require pynput in headless mode. It's not _that_ big of a deal, we can figure that out on the long term. But for now, we definitely need to not run those in headless.

modules/browser_object_navigation.py
ben-c-at-moz: Well, the object model is incomplete (by necessity—there's a lot of ground to cover here), and these selectors may be useful to future development of other functions. I will move them up, though, on second examination leaving them at the end doesn't help signify anything at all...

modules/browser_object_navigation.py
ben-c-at-moz: This is specifically because I don't want to send another `click` to the bar.

modules/browser_object_navigation.py
ben-c-at-moz: This was done before I started "fluentifying" my object model, I'll change it.

conftest.py
ben-c-at-moz: I'll go with something else. I don't like to use _ if I'm going to return the object.

modules/browser_object_navigation.py
ben-c-at-moz: I'm considering major parts of the chrome as Pages because otherwise I'll need to model the entire chrome as a single Page which I don't really want to do. I didn't __init__ because I wasn't doing anything to the Page object that required it. Inheritance from PyPOM's constructor was good enough.

modules/browser_object_navigation.py
ben-c-at-moz: Oh wait, no there was a reason—I will be returning an object (self) from inside these functions, and if I return that from within the Context Manager, Selenium gets amnesia about elements stored as class attributes and can no longer interact with them. I'll look at other ways to control the context here in future PRs.

modules/page_object_about_prefs.py
ben-c-at-moz: I updated the function names so that the call can be `about_prefs.dropdown(selector).function()` or `about_prefs.search_engine_dropdown().function()`. I can't set it as a class property of `AboutPrefs` because some prefs views have multiple dropdowns.

tests/credential/conftest.py
ben-c-at-moz: Good call, will add the dep and convert to JSON files in the future to avoid Common SDET Problem number one zillion: Bunch of test data embedded in the tests.

ben-c-at-moz:
> Added some thoughts. I also noticed you don't initialize all the pages, could you explain why?

Some of them didn't need anything added and the Page class supplies its own constructor.

modules/browser_object_navigation.py
jrbenny35: I guess what I am saying is to just enable chrome context when needed instead of switching to/from chrome. But this isn't a blocker, more so just curious on your thoughts.

jrbenny35 (approved):
Looks good! I would still push towards having a base page for chrome and content pages but that's more of an opinion. 🚀

ben-c-at-moz:
@jrbenny35 I'll keep in mind the idea to make the whole chrome context a Page—and if it comes up in future, I won't hesitate to implement it.

PR #8, opened 2024-04-29
Bc/fix certain broken shadow doms

ben-c-at-moz:
We locate elements of a Shadow DOM by attempting to use the Selenium `shadow_root` attribute. If that doesn't work, we use JS to get `.shadowRoot.children`. When we do the latter, and the element being searched is a direct child of the shadow root, selection failed because `WebElement.find_elements` only looks for the children of an element; it won't match the WebElement on which it is executed. We get around that with some more JS. This required additional typing for clarity.

Logging is added.

### DIFF
--- a/modules/page_base.py
+++ b/modules/page_base.py
@@ -1,4 +1,5 @@
 import json
+import logging
 import re
 
 from pypom import Page
@@ -7,7 +8,6 @@ from selenium.webdriver.remote.webelement import WebElement
 from selenium.webdriver.support import expected_conditions as EC
 
 from modules.util import PomUtils
-import logging
 
 # Convert "strategy" from the components json to Selenium By vals
 STRATEGY_MAP = {
--- a/modules/page_base.py
+++ b/modules/page_base.py
@@ -7,6 +7,7 @@ from selenium.webdriver.remote.webelement import WebElement
 from selenium.webdriver.support import expected_conditions as EC
 
 from modules.util import PomUtils
+import logging
 
 # Convert "strategy" from the components json to Selenium By vals
 STRATEGY_MAP = {
@@ -53,6 +54,8 @@ class BasePage(Page):
 
         # JSON files should be labelled with snake_cased versions of the Class name
         qualname = self.__class__.__qualname__
+        logging.info("======")
+        logging.info(f"Loading POM for {qualname}...")
         manifest_name = qualname[0].lower()
         for char in qualname[1:]:
             if char == char.lower():
@@ -68,6 +71,7 @@ class BasePage(Page):
 
     def load_element_manifest(self, manifest_loc):
         """Populate self.elements with the parse of the elements JSON"""
+        logging.info(f"Loading element manifest: {manifest_loc}")
         with open(manifest_loc) as fh:
             self.elements = json.load(fh)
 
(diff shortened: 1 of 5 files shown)


modules/page_base.py
```diff
@@ -131,17 +137,22 @@ def get_element(self, name: str, *label) -> WebElement:
...
         selector = self.get_selector(name, *label)
         if "shadowParent" in element_data:
+            logging.info(f"Found shadow parent {element_data['shadowParent']}...")
             shadow_parent = self.get_element(element_data["shadowParent"])
             shadow_element = self.utils.find_shadow_element(shadow_parent, selector)
             self.elements[name]["seleniumObject"] = shadow_element
             return shadow_element
         found_element = self.driver.find_element(*selector)
         self.elements[name]["seleniumObject"] = found_element
+        logging.info(f"Returning element {name}.\n")
```
pdehaan, line 155: This was the only one with a trailing newline (not that there is anything wrong/bad with that, just calling it out).

modules/page_object_about_prefs.py
```diff
@@ -1,5 +1,7 @@
 from pypom import Region
+from selenium.webdriver import Firefox
 from selenium.webdriver.common.by import By
+from selenium.webdriver.common.keys import Keys
```
pdehaan, line 4: Q: why the new imports? They don't _appear_ to be used, unless I'm missing some trick (disclaimer: I'm pretty Selenium/Python ignorant).

modules/page_base.py
ben-c-at-moz: Since this is the end of the selector-grabbing phase, the newline is there to separate each phase.

modules/page_object_about_prefs.py
ben-c-at-moz: My bad, will fix

modules/util.py
```diff
@@ -124,39 +127,78 @@ class PomUtils:
...
-            shadow_root = element.shadow_root
-            return [shadow_root]
-        except InvalidArgumentException:
+        logging.info(f"Getting shadow nodes from root {element}")
+
+        def shadow_from_script():
             shadow_children = self.driver.execute_script(
                 "return arguments[0].shadowRoot.children", element
             )
             if len(shadow_children) and any(shadow_children):
```
jrbenny35, line 142: I don't quite understand this logic. Could you explain why ```any(shadow_children)``` is needed?

modules/util.py
```diff
@@ -124,39 +127,78 @@ class PomUtils:
...
+        except InvalidArgumentException:
+            logging.info("Selenium shadow nav failed.")
+            return shadow_from_script()
         return []
 
+    def css_selector_matches_element(
+        self, element: Union[WebElement, ShadowRoot], selector: list
+    ) -> bool:
+        if type(element) == ShadowRoot:
+            return False
```
jrbenny35, line 165: Could do 
```suggestion
        if element is ShadowRoot:
            return False
```

jrbenny35 (approved):
Just had 2 thoughts

modules/util.py
ben-c-at-moz: When Selenium parses the results of the Javascript, it could be that any given child node is null, meaning we could have a list like `[None, None]` which I don't think is falsy. Let me check that...yep, that's considered `True` in python

jrbenny35: A list of None's is true lol, I love python. I guess a room full of no one is still a room!

modules/util.py
ben-c-at-moz: I just committed this, and then it somehow broke a test? That is an odd fact, but a true one. Looks like Selenium didn't properly implement `id` for that type.

ben-c-at-moz:
Got an approval from @jrbenny35

PR #9, opened 2024-05-03
Bc/improved expect method

ben-c-at-moz:
* Move context methods from Browser Object Nav class to BasePage
* Add more expect methods to BasePage
* Add selectors
* Lint
* Change hello-world TC task to linting task


modules/page_base.py
```diff
@@ -64,6 +64,20 @@ def __init__(self, driver, **kwargs):
...
+
+    def ensure_chrome_context(self):
+        """Make sure the Selenium driver is using CONTEXT_CHROME"""
+        if self._xul_source_snippet not in self.driver.page_source:
+            self.driver.set_context(self.driver.CONTEXT_CHROME)
+
+    def resume_content_context(self):
+        """Make sure the Selenium driver is using CONTEXT_CONTENT"""
+        if self._xul_source_snippet in self.driver.page_source:
+            self.driver.set_context(self.driver.CONTEXT_CONTENT)
```
pdehaan, line 79: I know this was just copy-paste from above, but&hellip; why is one `ensure_*` and the other is _`resume-*`_ prefix?

ben-c-at-moz: This was because it was only used in a Browser Object model, where `CONTEXT_CHROME` would have been the preferred context, but now that it's in the BasePage, it might need renaming. Thanks!

tests/about_prefs/conftest.py
```diff
@@ -0,0 +1,12 @@
+import pytest
+
+
+@pytest.fixture()
+def suite_id():
+    return ("2241", "Preferences")
```
sarinali, line 6: What does this string 2241 represent? Maybe it should be defined as a constant so there are no magic numbers?

ben-c-at-moz: This is how we're providing the test suite number. Adding it as a pytest fixture is essentially the same as adding it as a constant.

pdehaan:
LGTM, but I'll let @jrbenny35 and/or @sarinali give the final 👍 and :shipit: .

modules/page_base.py
```diff
@@ -64,6 +64,20 @@ def __init__(self, driver, **kwargs):
...
 
+    _xul_source_snippet = (
+        'xmlns:xul="http://www.mozilla.org/keymaster/gatekeeper/there.is.only.xul"'
+    )
+
+    def set_chrome_context(self):
+        """Make sure the Selenium driver is using CONTEXT_CHROME"""
+        if self._xul_source_snippet not in self.driver.page_source:
+            self.driver.set_context(self.driver.CONTEXT_CHROME)
+
```
jrbenny35, line 75: Would be cool to make this a context manager so you don't have to set `content` context.

jrbenny35 (approved):
Just had one thought, but all looks good

modules/page_base.py
ben-c-at-moz: Yeah, the context manager is preferable, but there are places where I can't rely on it. I did go through and CM-ify any of the code where it wouldn't break the test.

PR #13, opened 2024-05-10
Enable and Disable Autofill

sarinali:
# Description

Completes the test https://testrail.stage.mozaws.net/index.php?/cases/view/122347&group_by=cases:section_id&group_order=asc&display_deleted_cases=0&group_id=33867 

Which checks if autofill enable/disable works as expected.

## Bugzilla bug ID 

**ID:** 1895417
**Link:** https://bugzilla.mozilla.org/show_bug.cgi?id=1895417

## Type of change

- [x] New Test
- [x] New POM

Finished the entire test and POM, added some utility functions and tried to adhere to best practice.

# Screenshots / Explanations

N/A

# Comments / Concerns

When specifying the locale, for some reason only the data generated by faker in the CA and US locales will save correctly. The test will fail for other locales. Due to this, at the moment we are limited to only US and CA fake data until this issue is resolved.


modules/util.py
```diff
@@ -28,6 +28,48 @@ def random_string(self, n: int) -> str:
         shuffle(chars)
         return "".join(chars[:n])
 
+    def write_html_content(self, file_name: str, driver: Firefox, write_chrome: bool):
+        """
+        Takes the driver, the desired file name and the flag write_chrome, when true this flag will log the
+        web contents of the Chrome in the <file_name>.html and the regular page contents when it is fales.
```
ben-c-at-moz, line 34: `chrome`

modules/util.py
ben-c-at-moz, line 1: I couldn't see where this was being used in the test, would you point it out to me?

sarinali: I removed this now because I found it redundant, I just used something else in the test!

modules/util.py
```diff
@@ -66,7 +123,7 @@ def find_clear_and_fill(self, element_tuple: Iterable, term: str):
 
         ...
 
-        Parameters
+        Attributes
```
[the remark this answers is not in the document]
sarinali: Nope

modules/classes/autofill_base.py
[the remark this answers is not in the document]
sarinali: I put it in utilities instead!

modules/autofill_object_popup.py
```diff
@@ -0,0 +1,26 @@
...
+
+from modules.page_base import BasePage
+
+
+class AutofillPopup(BasePage):
+    """
+    Page Object Model for the popups that have to do with autofill
+    """
+
+    URL_TEMPLATE = ""
```
ben-c-at-moz, line 11: I previously had trouble with a blank `URL_TEMPLATE` but looks like it's working for you!

modules/autofill_object_popup.py
ben-c-at-moz, line 1: This might be better as `browser_object_autofill_popup`, so we know it's a Browser Object Model rather than a Page Object Model.

modules/autofill_object_save_info.py
ben-c-at-moz, line 1: This might be better as `page_object_autofill_test_basic`, so we know it's a Page Object Model

modules/classes/autofill_base.py
ben-c-at-moz, line 1: This is fine, but you might find Python's builtin dataclass easier to use. Up to you!

modules/util.py
```diff
@@ -44,13 +159,28 @@ class BrowserActions:
     def __init__(self, driver: Firefox):
         self.driver = driver
 
+    def clear_and_fill_no_additional_keystroke(self, webelement: WebElement, term: str):
```
ben-c-at-moz, line 162: You could modify `clear_and_fill` with a `press_enter` attr with `True` as the default, but this works too. Again, whichever you want is ok.

ben-c-at-moz (approved):
Looks good, but I would look at these comments before you merge, some I think would make sense to change.

modules/util.py
sarinali: i am going to leave this one because there are other tests that use the original method

ben-c-at-moz (approved):
LGTM!

PR #17, opened 2024-05-13
Hy/training task

sv-hyacoub:
# Description

Created a test to search via “Amazon” one-off search button.
In Navigation class I created a new getter and setter for “Amazon” one-off search button, a new methods to click on the awesome bar and a new method to search via “Amazon” one-off search button.

# Comments / Concerns

* I tried to include all methods from above in one ”type_in_awesome_bar_via_amazon_one_off” but I wasn’t sure if’s a good practice to have to setters in one method.
* While trying to configure git on pyCharm some git folders (ex. Hooks, info) were added.


ben-c-at-moz:
Thanks for participating!

PR #28, opened 2024-05-22
Bc/disable flaky

ben-c-at-moz:
# Description

Attempts to deflake the credentials test have been not so great. Let's disable it and mark deflaking it as tech debt. It will still run in local, just not in CI.

## Type of change

- [ ] New Test
- [ ] New POM
- [x] Disable Test

We use a new mark: `pytest.mark.unstable` to skip it in CI.

# Comments / Concerns

If this doesn't stop the conftest from running, then we'll still have error-based failures. I intend to rerun CI for all failing tests after merge.

### DIFF
--- a/ci_pyproject.toml
+++ b/ci_pyproject.toml
@@ -5,12 +5,13 @@ log_cli_level = "warn"
 markers = [
     "audio: test is reliant on audio",
     "pynput: test uses pynput package",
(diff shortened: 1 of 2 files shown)


sarinali (approved):
lgtm, asked ab `.toml` files in dm alr

PR #29, opened 2024-05-22
Form Autofill Enable Disable cc

sarinali:
# Description

Verifies that the setting in about:settings for payment information makes sure there is no popup panel

## Bugzilla bug ID

**ID:** 1895895
**Link:** https://bugzilla.mozilla.org/show_bug.cgi?id=1895895

## Type of change

- [x] New Test

Finishes the ticket

# Screenshots / Explanations

N/A

# Comments / Concerns

N/A


ben-c-at-moz (approved):
Yep, this will pass CI because it doesn't have to run there *for now*. So that's ok, lgtm

PR #32, opened 2024-05-24
Tw/test open new tab

Tracy-Walker:
# Description
Added test cases for opening new via UI [+] button and via keyboard shortcut

## Bugzilla bug ID
**ID:**
[**1895907**](https://bugzilla.mozilla.org/show_bug.cgi?id=1895907)
and
**ID:**
[**1895897**](https://bugzilla.mozilla.org/show_bug.cgi?id=1895907)

## Type of change

- [X] New Tests
- [X] Keyboard shortcut for opening a new tab

Completes these test cases


modules/browser_object_tabbar.py
```diff
@@ -29,6 +30,18 @@ def new_tab_by_button(self) -> BasePage:
             self.get_element("newtab-button").click()
         return self
 
+    def new_tab_by_keys(self, sys_platform) -> BasePage:
+        """Use keyboard shortcut to open a new tab"""
+        if sys_platform == "Darwin":
+            ActionChains(self.driver).key_down(Keys.COMMAND).send_keys("t").key_up(
```
ben-c-at-moz, line 36: You have access to ActionChains through the `BasePage` model, so if you don't mind changing calls to `ActionChains` to `self.actions`, and not import ActionChains here, that would be great.

tests/tabs/test_open_new_tab.py
```diff
@@ -0,0 +1,29 @@
...
+
+def test_open_new_tab_plus(driver: Firefox):
+    """
+    C134453 - A new tab can be opened from the dedicated button ("+")
+    """
+    browser = TabBar(driver).open()
+    driver.get("about:robots")
+    browser.set_chrome_context()
+    browser.new_tab_by_button()
+    WebDriverWait(driver, 10).until(EC.title_contains("Mozilla Firefox"))
```
ben-c-at-moz, line 16: Use `browser.expect` instead of `WebDriverWait...until`

tests/tabs/test_open_new_tab.py
```diff
@@ -0,0 +1,29 @@
...
+    assert driver.title == "Mozilla Firefox"
```
ben-c-at-moz, line 17: So, what do you think about cutting this line since we have an implied assertion on the previous? I'm of two minds, would love your opinion...

ben-c-at-moz (changes requested):
Looks good, minor changes

ben-c-at-moz (approved):
This is good. PLEASE DON'T MERGE until I figure out what is going on with the sync.

PR #33, opened 2024-05-24
Bc/add keyring unlock

ben-c-at-moz:
# Description

Adds a fixture called `unlock_keyring` which can be requested in tests that need an unlocked keyring. If a keyring password exists on the OS under test, the user will be prompted in the command line. This should not cause a problem for CI, as we will not set a keyring password there.

Usage is just to add `unlock_keyring` to the signature of any test function that needs it.

- [x] Quality of life / deflake

# Comments / Concerns

Created the possible future need for secrets management. Not implemented for Linux or Windows.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -2,6 +2,7 @@ import logging
 import os
 import platform
 from typing import List, Tuple
+from subprocess import (check_call, CalledProcessError)
 
 import pytest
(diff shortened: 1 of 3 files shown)


ben-c-at-moz:
Doesn't seem to actually fix the problem, we'll have to come back to this.

PR #34, opened 2024-05-28
Add get_elements as a method of BasePage

ben-c-at-moz:
# Description

Add get_elements, update the get_element func sig and change all uses of get_element that use labels to match new usage. Deflake the themes test.

## Type of change

- [x] QOL, deflake

# Comments / Concerns

Have tested, but lmk if it breaks anything.


sarinali (approved):
lgtm!

PR #38, opened 2024-05-30
don't test when there are no new tests

ben-c-at-moz:
Emergency merge


ben-c-at-moz:
Emergency merge to prevent bad pre-commits

PR #39, opened 2024-05-30
Sl/delete cc profile

sarinali:
# Description

Checks that upon deleting a cc profile, it does not show up in the grid anymore

## Bugzilla bug ID

**ID:** 122391
**Link:** https://testrail.stage.mozaws.net/index.php?/cases/view/122391

## Type of change

- [x] New Test
- [x] New POM

Finishes

# Screenshots / Explanations

N/A

# Comments / Concerns

Worth noting that I did this by saving 2 profiles then deleting 1, I didn't verify the actual values in the profile. Admittedly I can do this with 1 profile and delete that but I ran into an issue where I am not too sure how to just verify something doesn't exist without timing out.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,8 @@ env
 *.zip
 *.gz
 *.7z
+test.sh
+file[0-9]*.txt
 
 # Byte-compiled / optimized / DLL files
 __pycache__/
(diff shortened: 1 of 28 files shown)


ben-c-at-moz (changes requested):
Getting labels errors--not sure why I'm not getting a merge conflict

ben-c-at-moz (approved):
Saw some potential flakiness but the test passes now...

PR #42, opened 2024-06-04
Add t-linux-wayland workers

bhearsum:
These workers use a new image that runs Wayland, and should have a GUI available to run Firefox with.


ben-c-at-moz (approved):
LGTM

PR #47, opened 2024-06-10
Edit CC Profile from about:prefs

sarinali:
# Description

This will save a profile and then modify it in about:prefs and check if the modification was successful. This will modify the card number, the exp date and the year. 

## Bugzilla bug ID

**ID:** 1895885
**Link:** https://bugzilla.mozilla.org/show_bug.cgi?id=1895885

## Type of change

- [x] New Test

Finishes it

# Screenshots / Explanations

https://tropical-roadrunner-d78.notion.site/Blockers-8111925c19dd44f8b0460bd6a10b2265?pvs=4 

relevant blockers related to this 

# Comments / Concerns

This test has a ~2% fail rate. This is due to the flaky nature of changing the date, specifically the month. After trying mulitple workarounds, there are still cases where it will just not retrieve the updated information. This is slightly concerning because in a dry run of ~50 iterations, the month attr fails around once. 

Ideally it would be nice to find the root of the issue but as of now, no luck : (

Related Bugzilla bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1901622

### DIFF
--- a/tests/form_autofill/test_edit_credit_card.py
+++ b/tests/form_autofill/test_edit_credit_card.py
@@ -1,11 +1,10 @@
 import json
 import logging
-from time import sleep
 
 import pytest
 from selenium.webdriver import Firefox
-from selenium.webdriver.support import expected_conditions as EC
 from selenium.webdriver.common.by import By
+from selenium.webdriver.support import expected_conditions as EC
 
 from modules.browser_object import Navigation
 from modules.browser_object_autofill_popup import AutofillPopup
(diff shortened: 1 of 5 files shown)


tests/form_autofill/test_edit_credit_card.py
```diff
@@ -0,0 +1,129 @@
...
+
+from modules.browser_object import Navigation
+from modules.browser_object_autofill_popup import AutofillPopup
+from modules.page_object import AboutPrefs
+from modules.page_object_autofill_credit_card import CreditCardFill
+from modules.util import BrowserActions, Utilities
+
+tabs = [i for i in range(4)]
+
+
```
ben-c-at-moz, line 17: Would love to have `@pytest.mark.unstable` here, just because we have a known (though very low) amount of flake.

ben-c-at-moz (approved):
One note, but it looks good!

ben-c-at-moz (changes requested):
On second thought, we're really going to want that pytest marker, i'm finding it less stable in my tests than you did.

ben-c-at-moz (approved):
lgtm

PR #48, opened 2024-06-11
Bc/pocket setup

ben-c-at-moz:
# Description

Creates POM for about:newtab and implements a basic test of the layout. Because there are four tests that implement a layout check for about:newtab, this should "count" as four tests. Also adds some very basic selectors for about:config.

## Bugzilla bug ID (bugs forthcoming)

**ID:**
**Link:**

## Type of change

- [x] New Test
- [x] New POM

Completes the UI automation portion of C559390, C408037, C469597, and C408038. *This does not implement collection of DE, enGB, or FR localized builds of Firefox.* This isn't tech debt, it's just something we haven't implemented yet. We'll get there.

# Comments / Concerns

Sometimes, for reasons I still do not understand, we see 7 top sites instead of 8. I would love to figure out what the deal is with this.


flake_test.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
sarinali, line 1: `pytest --run-headless tests/form_autofill/test_edit_credit_card.py >> err_log.txt 2>&1` for logging would be helpful, esp if you have 10+ runs where the terminal would not keep all the logging history, also worth excluding incident tests? `pytest -m "not incident"` (Tracy found this one!)

modules/page_object_about_newtab.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def check_layout(self) -> BasePage:
+        """
+        Main method to check about:newtab layout is as expected
+        """
+        self.element_exists("incontent-search-input")
+        logging.info("Search bar exists")
+        self.element_exists("recent-activity-section")
+
+        for i in range(1, 22):
```
sarinali, line 111: why specifically 22 times? it would be nice to define this as a constant at the top for less _magical numberesque_ ness if you get me

sarinali (dismissed):
minor stuff, lgtm

modules/data/about_newtab.components.json
```diff
@@ -0,0 +1,79 @@
...
+        "groups": [
+            "requiredForPage"
+        ]
+    },
+
+    "article-description": {
+        "selectorData": "div[class='meta']",
+        "strategy": "css",
+        "groups": [
+            "articleTile"
```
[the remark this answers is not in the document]
ben-c-at-moz: You know, I'll remove it, I'd like to add some group-related functionality to the BasePage first.

modules/page_object_about_newtab.py
ben-c-at-moz: Will demystify this.

sarinali (approved):
lgtm!

PR #50, opened 2024-06-11
Active Tab 

sarinali:
# Description

This will check if the current tab is highlighted, this is done by checking the `visibility` attribute. Upon selecting tab, only the selected tab will have this attribute while the others do not. Also verified that selected tabs have the css property 'color': rgb(12, 12, 13) while the unselected ones are 'color': rgb(21, 20, 26). So we verify that visibility prop is there and also the colours match up.

## Bugzilla bug ID

**ID:** 1895904
**Link:** https://bugzilla.mozilla.org/show_bug.cgi?id=1895904

## Type of change

- [x] New Test

finishes

# Screenshots / Explanations

N/A

# Comments / Concerns

I hope this method of verification is satisfactory!

### DIFF
--- a/tests/form_autofill/test_clear_form.py
+++ b/tests/form_autofill/test_clear_form.py
@@ -1,5 +1,4 @@
 import pytest
-from selenium.common.exceptions import NoSuchElementException
 from selenium.webdriver import Firefox
 
 from modules.browser_object_autofill_popup import AutofillPopup
--- a/tests/tabs/test_active_tab.py
+++ b/tests/tabs/test_active_tab.py
@@ -1,0 +1,23 @@
+from selenium.webdriver import Firefox
+
+from modules.browser_object import TabBar
+
+
+def test_active_test(driver: Firefox):
+    """
+    C134646, ensures that the selected tab is highlighted
+    """
+    tabs = TabBar(driver).open()
+    num_tabs = 5
+
+    # opening 5 tabs
+    for i in range(num_tabs):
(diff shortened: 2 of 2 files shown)


ben-c-at-moz (changes requested):
Sadly, the test fails in dark mode: `AssertionError: assert ('rgb(251, 251, 254)' == 'rgb(12, 12, 13)'`

I also think it should be called `test_active_tab` rather than `test_active_test`.

sarinali:
> Sadly, the test fails in dark mode: `AssertionError: assert ('rgb(251, 251, 254)' == 'rgb(12, 12, 13)'`
> 
> I also think it should be called `test_active_tab` rather than `test_active_test`.

i had a lapse in naming capabilities earlier

sarinali:
> Sadly, the test fails in dark mode: `AssertionError: assert ('rgb(251, 251, 254)' == 'rgb(12, 12, 13)'`
> 
> I also think it should be called `test_active_tab` rather than `test_active_test`.

ok so i removed the css value checking, the reason for this is because i went and investigated the colour it should be in dark mode, the colour does not change. in fact, the css between selected and unselected is the same for the tab item, i verified that the `visuallyselected` field works by doing some experimental code checking if an unselected tab has that prop. it does not. so i think this is a valid way to verify. what do you think?

ben-c-at-moz (approved):
I do think the attribute is a fine way to do the test. Ready to merge.

PR #51, opened 2024-06-12
Pin Tab

sarinali:
# Description

This verifies that a pinned tab is pinned, non pinned tabs are unpinned. Then, it unpins the tab and verifies that it was unpinned.

## Bugzilla bug ID

**ID:** 1895903
**Link:** https://bugzilla.mozilla.org/show_bug.cgi?id=1895903

## Type of change

- [x] New Test
- [x] New POM

Finishes.

# Screenshots / Explanations

N/A

# Comments / Concerns

Context menu doesnt close but it doesn't affect anything else


ben-c-at-moz (changes requested):
So the test passes, but even in headless mode, the context menu appears and doesn't go away until I click somewhere, so this will hang any given CI. Let's figure out how to close these menus (without Pynput preferably)...

ben-c-at-moz (approved):
good to merge, no notes

PR #53, opened 2024-06-12
basic theme install test

ben-c-at-moz:
# Description

Creates a home for AMO POMs, as well as a POM for AMO Themes, adds POM elements to about:addons POM, completes test for C118174.

## Bugzilla bug ID

**Link:** https://bugzilla.mozilla.org/show_bug.cgi?id=1895913

## Type of change

- [x] New Test
- [x] New POM
- [x] Other Changes - BasePage add function to wait for URL to change

Completes test

# Screenshots / Explanations

Because the title in AMO's listings may not match the title of the theme when viewed in about:addons, had to test in a more robust, but less specific way (test that the theme has changed after install).

### DIFF
--- a/modules/data/about_addons.components.json
+++ b/modules/data/about_addons.components.json
@@ -21,5 +21,23 @@
         "selectorData": "back-button",
         "strategy": "class",
         "groups": []
+    },
+
+    "enabled-theme-title": {
+        "selectorData": ".theme-enabled-section .addon-name",
+        "strategy": "css",
+        "groups": []
+    },
+
+    "find-more-themes-button": {
+        "selectorData": "button[data-l10n-id='find-more-themes']",
+        "strategy": "css",
+        "groups": []
+    },
+
+    "install-theme-button": {
+        "selectorData": "button[data-l10n-id='install-theme-button']",
+        "strategy": "css",
+        "groups": []
     }
 }
--- a/modules/page_object_addons_mozilla_org.py
+++ b/modules/page_object_addons_mozilla_org.py
@@ -1,0 +1,26 @@
+from selenium.webdriver.common.by import By
+
+from modules.page_base import BasePage
+
+# PLEASE add all AMO page models to this file
+
+
+class AmoThemes(BasePage):
+    """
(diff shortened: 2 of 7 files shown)


modules/data/about_addons.components.json
```diff
@@ -21,5 +21,23 @@
...
         "groups": []
+    },
+
+    "enabled-theme-title": {
+        "selectorData": ".theme-enabled-section .addon-name",
+        "strategy": "css",
+        "groups": []
+    },
+
+    "find-more-themes-button": {
```
sarinali, line 32: nit picking up these two selectors can be made into 1 with the label, but up to u!

modules/page_object_addons_mozilla_org.py
```diff
@@ -0,0 +1,26 @@
...
+    def install_recommended_theme(self) -> BasePage:
+        """
+        Install the first recommended theme.
+        """
+        self.get_element("recommended-addon").click()
+        self.theme_title = self.get_element("theme-title").get_attribute("innerText")
+        self.get_element("install-button").click()
+        # TODO: Add toolbar browser object
+        with self.driver.context(self.driver.CONTEXT_CHROME):
+            self.driver.find_element(By.CSS_SELECTOR, "button[label='Add']").click()
```
sarinali, line 24: just curious, why was this done with the selenium way not the POM/BOM way

sarinali (approved):
lgtm, minor things but could be stylistic

modules/page_object_addons_mozilla_org.py
ben-c-at-moz: Because lazy. :) I didn't want to create a whole manifest file AND BOM file just to click two buttons. In the future, we'll need to make that BOM for sure, though.

modules/data/about_addons.components.json
ben-c-at-moz: They *could* but it's not like they're "good buddies" or anything. If they were two buttons that we always saw together, I would have, believe me. I love me some labels.

PR #56, opened 2024-06-12
Recently Closed Tabs

sarinali:
# Description

Ensures that the recently closed tabs function works in the panel menu

## Bugzilla bug ID

**ID:** 1895900
**Link:** https://bugzilla.mozilla.org/show_bug.cgi?id=1895900

## Type of change

- [x] New Test

Finishes

# Screenshots / Explanations

N/A

# Comments / Concerns

Maybe a random link generator would make this test more arbitrary in the future

### DIFF
--- a/modules/util.py
+++ b/modules/util.py
@@ -343,6 +343,8 @@ class PomUtils:
         The instance of WebDriver under test.
     """
 
+    allowed_selectors = set([By.ID, By.CLASS_NAME])
+
     def __init__(self, driver: Firefox):
         self.driver = driver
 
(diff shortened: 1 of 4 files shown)


modules/util.py
```diff
@@ -394,15 +396,20 @@ def find_shadow_chrome_element(
         self, nodes: list[WebElement], selector: list
     ) -> Union[WebElement, None]:
         logging.info("Selecting element in Chrome Context Shadow DOM...")
-        if selector[0] != By.ID:
+        if selector[0] not in self.allowed_selectors:
+            # if selector[0] != By.ID:
             raise ValueError(
-                "Currently shadow elements in chrome can only be selected by ID."
+                "Currently shadow elements in chrome can only be selected by ID and classname."
```
ben-c-at-moz, line 402: Didn't we create functionality for selecting shadow chrome elements by tag name? What happened to that code?

tests/tabs/test_recently_closed_tabs.py
```diff
@@ -0,0 +1,55 @@
+import logging
+
+from selenium.webdriver import Firefox
+from selenium.webdriver.support import expected_conditions as EC
+
+from modules.browser_object import PanelUi, TabBar
+
+links = [
```
ben-c-at-moz, line 8: I might consider only using sites where you know the load latency is very small. If you can confirm that is the case for all of these, go for it. If not, you can open a bunch of about pages instead.

ben-c-at-moz (changes requested):
These are small things, but I would actually love to see them addressed. Thanks!

modules/util.py
sarinali: i think i discarded the changes bc it wasn't relevant atm to the shadow element. in the end, it was for the pin tab PR and that was figured out with a workaround

ben-c-at-moz: OK, that's fine

PR #59, opened 2024-06-13
add hide popup

sarinali:
# Description

hide popup method

Credits to Ben the 🐐


ben-c-at-moz (approved):
There you go! Good job.

PR #61, opened 2024-06-13
Navigation Through Tabs

sarinali:
# Description

Verifies that the arrows on the tabs can navigate through tabs

## Bugzilla bug ID

**ID:** 1895901
**Link:** https://bugzilla.mozilla.org/show_bug.cgi?id=1895901

## Type of change

- [x] New Test

Finishes

# Screenshots / Explanations

N/A

# Comments / Concerns

It will just verify that the position increased/decreasing after pressing one of the buttons. This might not be optimal correctness.


modules/page_base.py
```diff
@@ -294,7 +294,7 @@ def element_exists(self, name: str, labels=[]) -> Page:
         )
         return self
 
-    def element_visible(self, name: str, labels=[]) -> Page:
+    def element_visible(self, name: str, *labels) -> Page:
```
ben-c-at-moz, line 297: don't want this reversion!

ben-c-at-moz (changes requested):
Fix the one reversion and you're good to go.

ben-c-at-moz (approved):
Good to merge!

modules/page_base.py
ben-c-at-moz: I did it!

PR #62, opened 2024-06-13
Write test to reopen tab through ctx menu

ben-c-at-moz:
# Description
This adds a test that closes a tab, then reopens it through the tab context menu.

## Bugzilla bug ID

**Link:** https://bugzilla.mozilla.org/show_bug.cgi?id=1895899

## Type of change

- [x] New Test

Completes test

### DIFF
--- a/modules/data/tab_context_menu.components.json
+++ b/modules/data/tab_context_menu.components.json
@@ -10,5 +10,11 @@
         "selectorData": "context_unpinTab",
         "strategy": "id",
         "groups": []
+    },
+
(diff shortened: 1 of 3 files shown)


ben-c-at-moz:
> Why are you using `with driver.context(driver.CONTEXT_CHROME):` instead of calling our methods `set_chrome_context()` and `set_content_context()`
> 
> otherwise looks good

I did this because we don't have a BOM for "fx window manager", nor could we really create one if we wanted to. Unfortunately for some of these tab tests, the cleanest way is some bare Selenium.

PR #65, opened 2024-06-14
Recently Closed Tabs Fixed

sarinali:
# Description

Had some **really** bad git issues in my previous branch (https://github.com/mozilla/fx-desktop-qa-automation/pull/56) so I just made a new one and this has all the same changes as the old one.

## Bugzilla bug ID

**ID:** 1895900
**Link:** https://bugzilla.mozilla.org/show_bug.cgi?id=1895900

## Type of change

-  [x] New Test

finishes and fixes

# Screenshots / Explanations

n/a

# Comments / Concerns

n/a

### DIFF
--- a/modules/util.py
+++ b/modules/util.py
@@ -343,7 +343,7 @@ class PomUtils:
         The instance of WebDriver under test.
     """
 
-    allowed_selectors = set([By.ID, By.CLASS_NAME])
+    allowed_selectors = set([By.ID, By.CLASS_NAME, By.TAG_NAME])
 
     def __init__(self, driver: Firefox):
         self.driver = driver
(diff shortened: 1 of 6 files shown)


modules/util.py
```diff
@@ -343,6 +343,8 @@ class PomUtils:
         The instance of WebDriver under test.
     """
 
+    allowed_selectors = set([By.ID, By.CLASS_NAME])
```
ben-c-at-moz, line 346: Since you pulled this out of the method, the name is no longer specific enough to describe what it does.

ben-c-at-moz: Also, since you left TAG_NAME in the method, it should be in this set too.

ben-c-at-moz (changes requested):
Minor change, but otherwise looks good

modules/util.py
sarinali: done! fixes are pushed

PR #66, opened 2024-06-15
test find toolbar search

ben-c-at-moz:
# Description

- Add BOM for Find Toolbar
- Add Find Toolbar suite 
- Add linked test
- update commit hook py script to emit errors properly
- Add BrowserActions method to get the color scheme of an element   

## Bugzilla bug ID

**Link:** https://bugzilla.mozilla.org/show_bug.cgi?id=1902751

## Type of change

- [x] New Test
- [x] New POM
- [x] Commit hook improvement, method to get element colorscheme

Complete

# Screenshots / Explanations

We use `pillow` to do visual comparison.

# Comments / Concerns

New dependency: `pillow`.


conftest.py
```diff
@@ -185,14 +185,23 @@ def driver(
...
 
 @pytest.fixture()
-def screenshot(driver: webdriver.Firefox, opt_ci: bool):
-    def _screenshot(filename):
+def screenshot(driver: webdriver.Firefox, opt_ci: bool) -> Callable:
+    """
+    Factory fixture that returns a screenshot function.
+    """
+
+    def _screenshot(filename: str) -> str:
```
sarinali, line 193: curious as to why you have a private function defined in the function with the same name? what was the design decision behind putting everything in the same function scope?

modules/browser_object_tabbar.py
```diff
@@ -243,8 +243,6 @@ def wait_for_num_tabs(self, num_tabs: int) -> BasePage:
         Waits for the driver.window_handles to be updated accordingly with the number of tabs requested
         """
         try:
-            self.wait.until(
```
sarinali, line 246: is this an auto formatter? each time i do the ruff thing it makes it 3 lines, is that wrong im just curious

sarinali (commented):
minor comments but otherwise ok

conftest.py
ben-c-at-moz: This is a common PyTest fixture pattern called a "function factory" or "factory fixture". It's the way you add a fixture that refers to a function itself, so that you can request that fixture and receive a callable that you can later execute with arguments.

modules/browser_object_tabbar.py
ben-c-at-moz: This is definitely an autoformatter, and I think ours are fighting. Let's get on the same page with Ruff version etc.

sarinali (approved):
lgtm now!

PR #67, opened 2024-06-17
Bc/autofill in private

ben-c-at-moz:
# Description

Tests that autofill data is not saved in private mode. NB: The test case describes two different tests, this test only executes the _second_ one, and the TC should be divided to conform to best practices.

## Bugzilla bug ID

**Link:** https://bugzilla.mozilla.org/show_bug.cgi?id=1895883

## Type of change

- [x] New Test
- [x] Other Changes: QoL improvement on BasePage, test element does not exist

Completes the test

# Comments / Concerns

The first half of the written TC is difficult to follow. If we'd like to write that test as automation, we should divide the TC and re-triage.


sarinali (approved):
lgtm

PR #68, opened 2024-06-17
Setting Default Profile

sarinali:
# Description

Verifies that the profile shows as default when it is set as the default profile.

## Bugzilla bug ID

**ID:** 1902744 
**Link:** https://bugzilla.mozilla.org/show_bug.cgi?id=1902744

## Type of change

- [x] New Test
- [x] New POM
- [x] Other Changes: Modified `page_base.py` to accept an additional parameter `parent_element` which would ensure that 

Done.

# Screenshots / Explanations

N/A

# Comments / Concerns

Same 6 fail from Nightly change but these were fixed in another PR.


tests/profile/test_set_default_profile.py
```diff
@@ -0,0 +1,87 @@
...
+    # verify that some profile is the default
+    profile_header_in_use = about_profiles.get_element(
+        "profile-container-item-default-header"
+    )
+    assert profile_header_in_use is not None
+
+    cur_default = -1
+
+    # find index that is the current default
+    for i in range(len(profiles)):
```
ben-c-at-moz, line 29: Iterating is fine, but do we need to do it? All we need to know is that there _is_ a default profile, right? Would it be quicker to just find that element anywhere in the `about:profiles` page?

tests/profile/test_set_default_profile.py
```diff
@@ -0,0 +1,87 @@
...
+            cur_default = i
+            break
+
+    # no default profile could be found
+    if cur_default == -1:
+        logging.warn("Could not find a currently active default profile.")
+        assert False
+
+    # select a non default profile randomly
+    profile_index = random.randint(0, len(profiles) - 1)
```
ben-c-at-moz, line 53: could make a set of all the profiles that are not the current, then choose a random from that, would avoid iteration.

ben-c-at-moz (approved):
A couple of dorky notes from me, but the code works and looks fine.

tests/profile/test_set_default_profile.py
sarinali: iteration is required to find the default profile!

tests/profile/test_set_default_profile.py
sarinali: Got it. Fix was just pushed

ben-c-at-moz (approved):
Nice work!

PR #70, opened 2024-06-18
As/search suggestion for engine selection

soncuteanca:
# Description

This verifies that when entering "@" in awesome bar, a list of search engines is suggested, and upon selecting an option, the search is performed using the selected engine.

## Bugzilla bug ID

**ID: 1902767**
**Link: https://bugzilla.mozilla.org/show_bug.cgi?id=1902767**

## Type of change

- [ x] New Test

Finished

# Screenshots / Explanations

# Comments / Concerns


tests/address_bar_and_search/test_suggestion_engine_selection.py
```diff
@@ -0,0 +1,47 @@
...
+    nav = Navigation(driver).open()
+    nav.type_in_awesome_bar("@")
+
+    # Use base selector to find all search engine suggestions
+    suggestion_list_items = nav.get_element("search-suggestion-list", multiple=True)
+
+    # Filter elements to find the one that matches the desired site
+    suggestion_list_item = None
+    for item in suggestion_list_items:
+        try:
```
ben-c-at-moz, line 26: You might be able to avoid this if you add `"doNotCache"` to the element's groups.

modules/data/navigation.components.json
```diff
@@ -100,6 +100,11 @@
         "selectorData": "nav-bar",
         "strategy": "id",
         "groups": []
-    }
+    },
 
+    "search-suggestion-list": {
+        "selectorData": "div.urlbarView-row[type='search_engine'] span.urlbarView-title",
+        "strategy": "css",
+        "groups": []
```
ben-c-at-moz, line 108: Consider adding `"doNotCache"` to groups to avoid stale elements

ben-c-at-moz (changes requested):
One thing to try, but it looks good!

tests/address_bar_and_search/test_suggestion_engine_selection.py
soncuteanca: I did a second commit, with the "doNotCache" added to the element's groups, but intermittently i run into the same staleElement problem. Is there anything I could do on my current commit, or I should use a try-except block afterall?

ben-c-at-moz:
@soncuteanca I'm still getting a failure on the test when trying for Amazon.com -- perhaps it's actually looking for just "Amazon"? Not sure...

tests/address_bar_and_search/test_suggestion_engine_selection.py
```diff
@@ -0,0 +1,34 @@
...
+def test_search_suggestion_for_engine_selection(driver: Firefox, site: str):
+    """
+    C1365228 - Test to verify that when entering "@", a list of search engines is suggested, and upon selecting an
+    engine, the search is performed using the selected engine.
+    """
+    nav = Navigation(driver).open()
+    nav.type_in_awesome_bar("@")
+
+    # Use base selector to find all search engine suggestions
+    suggestion_list_items = nav.get_element("search-suggestion-list", multiple=True)
```
ben-c-at-moz, line 20: Recomment `nav.get_elements(...` instead. (And then you can leave off the `multiple=True`)

tests/address_bar_and_search/test_suggestion_engine_selection.py
```diff
@@ -0,0 +1,34 @@
...
+@pytest.mark.parametrize("site", sites)
+def test_search_suggestion_for_engine_selection(driver: Firefox, site: str):
+    """
+    C1365228 - Test to verify that when entering "@", a list of search engines is suggested, and upon selecting an
+    engine, the search is performed using the selected engine.
+    """
+    nav = Navigation(driver).open()
+    nav.type_in_awesome_bar("@")
+
+    # Use base selector to find all search engine suggestions
```
ben-c-at-moz, line 19: So, we're having flaky tests runs here, probably because the menu is taking a bit of time to populate. I recommend using some kind of iteration to `get_element` each of the target search engines individually before moving on. This will help because selenium `find_elements` doesn't wait for all possible matches to appear, it just returns when ANY elements match the selector. You could then _not_ use `get_elements` in the next line (if you collect the elements during iteration) OR you could keep that if that's your preferred way of collecting the elements.

ben-c-at-moz (changes requested):

tests/address_bar_and_search/test_suggestion_engine_selection.py
soncuteanca: I've implemented a retry mechanism for locating search suggestion elements, not sure this is the best approach, I've run my tests several times without any failure though. Let me know if I should try a different approach on this one.

ben-c-at-moz: This is still flaky for me (though quite a bit better than before).  You might try adding a new selector or changing the existing selector so that you can get the `data-l10n-args` attribute on the `.urlbarView-title` element. Presumably this command (don't use it exactly as written of course) would wait for your element to exist:
`nav.driver.find_element(By.CSS_SELECTOR, f'.urlbarView-title[data-l10n-args="{\"engine\":\"{site}\"}]')`

But that is pretty complex. You could change the check inside your iterator to
```python
nav.expect(
  EC.text_to_be_present_in_element(
    nav.get_element("search-suggestion-list"),
    site
  )
)
```

ben-c-at-moz (changes requested):
I'm sorry, it's still flaky for me...

ben-c-at-moz (approved):
I just added a little something and now it works without hard waits or iterative checks

soncuteanca:
With the latest changes, I have achieved a 15/15 pass rate locally. Is there anything else I should do here?

ben-c-at-moz:
@soncuteanca No, you just needed to resolve all the conversations by pressing the "Resolve Conversation". It's good to merge now

PR #74, opened 2024-06-21
hy/google search counts us

sv-hyacoub:
# Description

This test ensures that search count telemetry data is collected accordingly for Google.

## Bugzilla bug ID

**ID: 1902774**
**Link: https://bugzilla.mozilla.org/show_bug.cgi?id=1902774**

## Type of change

- [x] New Test
- [x] New POM

Completed.

# Screenshots / Explanations

N/A.

# Comments / Concerns

Created a POM for about:telemetry with a function to decode url to base64
Created a function in BasePage to switch tab.
Created a test for google search count, I used jsonpath-ng 1.6.1 to parse the json.
I had to use time.sleep() after the search otherwise both of the telemetries weren't displayed.


tests/address_bar_and_search/report.html
ben-c-at-moz, line 1: This report shouldn't be in the PR... not sure how it's here given the line in .gitignore that should eliminate it. Could you `git rm` this file? I'll try to fix the .gitignore in a future PR

tests/address_bar_and_search/test_google_search_counts_us.py
```diff
@@ -0,0 +1,32 @@
+import time
+
+from selenium.webdriver import Firefox
+
+from modules.browser_object_about_telemetry import AboutTelemetry
+from modules.browser_object_navigation import Navigation
+from jsonpath_ng import parse
```
ben-c-at-moz, line 7: Because this package doesn't exist in CI, we're getting failures. Run `pipenv install jsonpath_ng` to fix this. I'll need to do a brief review of the package for size and security, but I think it'll be fine.

ben-c-at-moz (changes requested):
A couple of odd things but the test itself looks good from here.

ben-c-at-moz:
One more thing, I'm going to need you to merge main before I can approve the PR

tests/address_bar_and_search/report.html
sv-hyacoub: Done, I eliminated that file.

tests/address_bar_and_search/test_google_search_counts_us.py
sv-hyacoub: Done.

ben-c-at-moz (approved):
lgtm

PR #79, opened 2024-06-25
Frequently Used Context Actions

sarinali:
# Description

Verifies that "Save Page As..", "Take Screenshot" and "Inspect" options all work from the context menu.

## Bugzilla bug ID

**ID:** 1903797
**Link:** https://bugzilla.mozilla.org/show_bug.cgi?id=1903797

## Type of change

- [x] New Test
- [x] New POM

Finishes

# Screenshots / Explanations

N/A

# Comments / Concerns

Again, like the image context PR, this one uses pynput and saving files. This is sometimes unreliable because of the system dialog not showing up in time. A hardcoded sleep of 0.5 seconds is put in the source code. For now it was marked as unstable.


tests/menus/test_frequently_used_context_menu.py
```diff
@@ -0,0 +1,120 @@
...
+    nav = Navigation(driver)
+    util = Utilities()
+    ba = BrowserActions(driver)
+
+    # right click something that is not a hyperlink
+    title_header = example_page.get_element("title-header")
+    context_menu.context_click_element(title_header)
+
+    with driver.context(driver.CONTEXT_CHROME):
+        save_page_as = context_menu.get_context_item("context-menu-save-page-as")
```
ben-c-at-moz, line 34: It seems weird that we have to explicitly call change the context here, is there a way to push that functionality into the BOM method `get_context_item`?

tests/menus/test_frequently_used_context_menu.py
```diff
@@ -0,0 +1,120 @@
...
+        ba.key_press_release(Key.tab)
+
+        ba.key_press_release(Key.tab)
+
+    # Press and release the Enter key
+    ba.key_press_release(Key.enter)
+
+    # Wait for the animation to complete
+    with driver.context(driver.CONTEXT_CHROME):
+        nav.wait_for_download_animation_finish(downloads_button)
```
ben-c-at-moz, line 64: Same as above, is there a way to make this method change the context for us?

tests/menus/test_frequently_used_context_menu.py
```diff
@@ -0,0 +1,120 @@
...
+    C2637623.2: take screenshot works from context menu
+    """
+    # create objects
+    context_menu = ContextMenu(driver)
+    driver.get("https://example.com")
+    example_page = ExamplePage(driver)
+
+    # ensure that the screenshot is not present
+    with driver.context(driver.CONTEXT_CHROME):
+        example_page.element_does_not_exist("take-screenshot-box")
```
ben-c-at-moz, line 85: I think I've refactored this function recently and it doesn't require explicit context switching anymore. If not, I think you could do that!

ben-c-at-moz (dismissed):
Minor things, if you want to tackle them that would be great. If not, the tests work as is.

tests/menus/test_frequently_used_context_menu.py
sarinali:
```def element_does_not_exist(self, name: str, labels=[]) -> Page:
        """Expect helper: wait until element exists or timeout"""
        original_timeout = self.driver.timeouts.implicit_wait
        self.driver.implicitly_wait(0)
        self.instawait.until_not(
            EC.presence_of_all_elements_located(self.get_selector(name, labels=labels))
        )
        self.driver.implicitly_wait(original_timeout)
        return self``

Sorry if this is a stupid question, where is the chrome context searching taking place in this function?

tests/menus/test_frequently_used_context_menu.py
sarinali: Got it!

tests/menus/test_frequently_used_context_menu.py
sarinali: I added it in!

tests/menus/test_frequently_used_context_menu.py
sarinali: Ok, I'll wait a moment. I'm not getting any changes for this function hm

ben-c-at-moz (approved):
lgtm

PR #80, opened 2024-06-25
fix multiple clicks

ben-c-at-moz:
# Description

- Create a function in BasePage that emits _n_ number of clicks in succession to an element
- Fix spelling issues in a test
- Improve context handling in multiple clicks

## Type of change

- [x] Spelling, QoL (BasePage ergonomics)

# Screenshots / Explanations

The multi_click function allows us to send an arbitrary _n_ clicks to an element, in the POM's native context, whether described by a POM element name, a Selenium selector tuple, or a WebElement

# Comments / Concerns

This needs to be front and center when multi-clicking at n > 3


ben-c-at-moz:
> lgtm, though I know this is going to break the tests I just created, as they rely on triple_click_element(). Land this, after you resolved the straight forward conflict, then I'll deal with the tests it breaks in my next PR.

I will actually just fix that in this PR

PR #82, opened 2024-06-26
As/addon suggestion based on search input

soncuteanca:
# Description

This test ensures that a matching add-on suggestion appears based on the search input and that clicking on the suggested add-on redirects to the corresponding add-on page.

## Bugzilla bug ID

**ID: 1902787**
**Link: https://bugzilla.mozilla.org/show_bug.cgi?id=1902787**

## Type of change

- [x ] New Test

In progress

# Screenshots / Explanations

# Comments / Concerns
I have a passing test, but I need to work on optimizing the timing. None of the waiting condition approach I've tried so far did the trick.

### DIFF
--- a/ci_pyproject.toml
+++ b/ci_pyproject.toml
@@ -7,6 +7,7 @@ markers = [
     "pynput: test uses pynput package",
     "incident: incident smoke tests",
     "unstable: temporary mark for unstable tests",
+    "slow: test is clocked at more than 30s on modern machines",
     "ci: basic tests to run in ci",
     "locale_de: tests run in DE locale versions",
     "locale_fr: tests run in FR locale versions",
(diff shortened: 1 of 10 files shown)


ben-c-at-moz:
I made some suggested changes (and one BOM method that turned out not to be helpful this time but may be helpful in future), in the branch bc/amendments-to-82. @Tracy-Walker pointed out to me that this might be a good time not to parametrize since we have to wait every time the browser starts. I also made it wait only as long as it needed to before proceeding (sort of).

ben-c-at-moz (changes requested):
If you get a moment to de-parametrize this, I think that would make it nicer for us to run. Also, I think you only need to wait 10 seconds after clicking the url bar, but that may be wrong. If you try it and it doesn't work, that's fine.

soncuteanca:
I've de-parameterized the test. However, I have poor internet connection since yesterday, the test only completes successfully in 40 seconds. As a result, I need both time.sleep statements to remain in place.

ben-c-at-moz (approved):
I've added a marker to the test to indicate that it's slow, and we'll come back to it when we optimize tests later.

PR #86, opened 2024-07-02
hy/test SAP google adclick

sv-hyacoub:
# Description

This test ensures that adclick data is collected accordingly for Google.

## Bugzilla bug ID

**ID: 1902777**
**Link: https://bugzilla.mozilla.org/show_bug.cgi?id=1902777**

## Type of change

- [x] New Test

Completed.

# Screenshots / Explanations

N/A.

# Comments / Concerns

I'm using time.sleep before clicking on "Accept cookies" otherwise it can't find the element, I tried using element_exists or nav.expect(ec.title_contains("Google Search")) but that didn't help.


modules/data/navigation.components.json
```diff
@@ -146,5 +146,17 @@
         "groups": [
             "doNotCache"
         ]
+    },
+
+    "accept-cookies": {
+    "selectorData": "L2AGLb",
```
ben-c-at-moz, line 152: Generally, we don't want to select on ids or classes that look like this--often a base64 number or hash. These are generated by various web frameworks and can vary from version to version of the website, or even by the session. I suspect this is why it is failing in CI.

ben-c-at-moz (changes requested):

modules/data/navigation.components.json
emilghittasv: Good point. Hmm I can't find a better strategy for that element other than it's visible text.

Can we change the strategy to xpath and the selectorData to "//div[text()='Accept all']" ? It should work.

sv-hyacoub: I just made a commit with all the requested changes.

emilghittasv (dismissed):
lgtm

emilghittasv (approved):

PR #87, opened 2024-07-02
Intervention Card Firefox

sarinali:
# Description

Finishes verifying elements of the intervention card in firefox

## Bugzilla bug ID

**ID:** 1902765
**Link:** https://bugzilla.mozilla.org/show_bug.cgi?id=1902765

## Type of change

- [x] New Test

Finishes

# Screenshots / Explanations

N/a

# Comments / Concerns

1. In selenium, the behaviour is similar to private browsing mode so only 2 results show up instead of the typical 8 with the search suggestions.
2. Added a helper method that gets all children of a webelement. Moved hover element into page base for support in all BOM/POMs


ben-c-at-moz (approved):
Looks good!

PR #91, opened 2024-07-03
Deleting Profiles

sarinali:
# Description

Verifies that you can delete a profile. Currently only checks that the "don't delete files" option works

## Bugzilla bug ID

**ID:** 1902746
**Link:** https://bugzilla.mozilla.org/show_bug.cgi?id=1902746

## Type of change

- [x] New Test
- [x] Other Changes: Lots of file utilities

Finishes

# Screenshots / Explanations

N/A

# Comments / Concerns

Did not verify that deleting files works, this is because there is no way to interact with that button using selenium alerts. Possibly in the future we would be able to?


modules/page_object_about_profiles.py
```diff
@@ -7,3 +10,58 @@ class AboutProfiles(BasePage):
...
+        #     self.system_dialog_interation_windows()
+        # else:
+        #     print("running")
+        #     self.system_dialog_interation_macos(profile_name, ba)
+        # sleep(10)
+
+    def system_dialog_interation_linux(self):
+        pass
+
+    def system_dialog_interation_macos(self, profile_name: str, ba: BrowserActions):
```
ben-c-at-moz, line 47: Um, I think this might be some leftover debugging stuff...

tests/profile/test_delete_profile.py
```diff
@@ -0,0 +1,149 @@
...
+        profiles = profile_container.find_elements(By.XPATH, "./child::*")
+        new_profile = None
+
+        # find the newly created profile
+        for profile in profiles:
+            profile_header = about_profiles.get_element(
+                "profile-container-item-profile-name", parent_element=profile
+            )
+            logging.info(
+                f"Current detected profile: {profile_header.get_attribute("innerHTML")}"
```
ben-c-at-moz, line 89: Need to change parenthesized content to `'innerHtml'`

ben-c-at-moz (changes requested):
A couple of must-fixes, and I'm somehow getting failures on `tests/menus/test_image_context_menu_actions.py` now? I don't know how they could be related, but they passed in another branch recently...

modules/page_object_about_profiles.py
sarinali: Yeah, forgot to delete!

tests/profile/test_delete_profile.py
sarinali: Got it!

PR #96, opened 2024-07-05
Vs/test default search provider can be changed

vsangereanMOZ:
Description
This test makes 3 checks : One that the "Change search settings" opens about:preferences#search, One that it opens a new tab for about:preferences#search when a site is loaded into the tab and that the default search engine can be changed and settings are also applied to the search bar

Bugzilla bug ID
ID: [1902768](https://bugzilla.mozilla.org/show_bug.cgi?id=1902768)
Link: [1365245](https://testrail.stage.mozaws.net/index.php?/cases/view/1365245)

Type of change

[x ] New Test
 New POM
 Other Changes (Please specify)
How does this resolve / make progress on that bug?
The test is completed

Screenshots / Explanations
N/A

Comments / Concerns
Needs approval


ben-c-at-moz (approved):
Looks like it works well! Good job!

emilghittasv (approved):
lgtm!

PR #108, opened 2024-07-11
Language Packs (about:addons)

sarinali:
# Description

Ensures that installing a language pack from about:addons displays correctly in settings and changes the locale in about:preferences.

## Bugzilla bug ID

**Testrail:** https://testrail.stage.mozaws.net/index.php?/cases/view/1549408
**Link:** https://bugzilla.mozilla.org/show_bug.cgi?id=1905206

## Type of change

- [x] New Test

N/A

# Screenshots / Explanations

N/A

# Comments / Concerns

Both tests were parameterized so that this can be easily extended to many languages later.


modules/data/generic_page.components.json
```diff
@@ -3,5 +3,43 @@
         "selectorData": "body",
         "strategy": "tag",
         "groups": []
+    },
```
ben-c-at-moz, line 6: Why are these in GenericPage? Are these elements in the Content Context? Kind of looks like no, but I could be wrong...

ben-c-at-moz: Yeah, I think these components and methods should be in a new manifest, `amo_language_packs.components.json` and you should probably make a new class in the `page_object_addons_mozilla_org.py` POM. I really think we'll be seeing this page again...

modules/page_object_about_prefs.py
```diff
@@ -58,8 +58,13 @@ def select_option(self, option_name: str):
             if len(matching_menuitems) == 0:
                 return False
             elif len(matching_menuitems) == 1:
-                matching_menuitems[0].click()
-                self.wait.until(EC.element_to_be_selected(matching_menuitems[0]))
+                if double_click:
+                    AboutPrefs(self.driver).double_click(
```
ben-c-at-moz, line 62: why is this not `self.double_click(`?

tests/language_packs/test_language_pack_install_addons.py
```diff
@@ -0,0 +1,69 @@
...
+
+@pytest.mark.parametrize("drop_down_name, shortform, localized_text", LANGUAGES)
+def test_language_pack_install_from_addons(
+    driver: Firefox, drop_down_name: str, shortform: str, localized_text: str
+):
+    """
+    C1549408: verify that installing a language pack from about:addons will correctly change the locale
+    """
+    # declaring objects and navigating
+    generic_page = GenericPage(driver)
```
ben-c-at-moz, line 18: See above, I think a POM for this page would be useful.

tests/language_packs/test_language_pack_install_addons.py
```diff
@@ -0,0 +1,69 @@
...
+    driver.get(LANGUAGE_ADDONS_LINK)
+
+    # ensuring the page was loaded
+    generic_page.custom_wait(timeout=20).until(
+        lambda _: generic_page.get_element("language-addons-title") is not None
+    )
+
+    # grab the appropriate link and wait until the page is loaded
+    language_row = generic_page.get_element("language-addons-row")
```
ben-c-at-moz, line 27: For many of these steps, they could be abstracted into methods on a new POM, but how many and which is a question I'll leave to you...

ben-c-at-moz (changes requested):
Right ideas, but I do have a couple of structural suggestions...

modules/page_object_about_prefs.py
sarinali: self refers to `Dropdown` class. So I have to create an instance of `AboutPrefs` (since the `double_click` method is not static) 

Relevant logs: (when i use `self`)

```
FAILED tests/language_packs/test_language_pack_install_addons.py::test_language_pack_install_from_addons[Italiano-it-Imposta alternative\u2026] - AttributeError: 'Dropdown' object has no attribute 'double_click'
```

modules/data/generic_page.components.json
sarinali: Sure, I will work on refactoring into a POM

ben-c-at-moz (approved):
Thanks for addressing those comments...looks great

PR #114, opened 2024-07-15
As/tls protocol

soncuteanca:
# Description

Verify that TLS v1.2 protocol is handled correctly and returns the proper information.

## Bugzilla bug ID

**Testrail: https://testrail.stage.mozaws.net/index.php?/cases/view/192739**
**Link: https://bugzilla.mozilla.org/show_bug.cgi?id=1905188**

## Type of change

- [x ] New Test

Completed

# Screenshots / Explanations

# Comments / Concerns


tests/security_and_privacy/test_tls_v1_2_protocol.py
```diff
@@ -0,0 +1,47 @@
...
+        nav.get_element("connection-secure-button").click()
+        nav.get_element("more-information-button").click()
+
+        driver.switch_to.window(driver.window_handles[-1])
+
+        technical_details = WebDriverWait(driver, 10).until(
+            EC.presence_of_element_located((By.ID, "security-technical-shortform"))
+        )
+
+        expected_technical_details = "Connection Encrypted (TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256, 128 bit keys, TLS 1.2)"
```
ben-c-at-moz, line 44: So for some reason, the info I get on my machine when I go to the site is:

`Connection Encrypted (TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384, 256 bit keys, TLS 1.2)`, the difference being `RSA` instead of `ECDSA`. We should figure out what's going on here before we merge.

soncuteanca: It looks like I've made an oversight. The string mentioned in the C192739 test case is indeed the one with RSA. I have verified this on two different test machines (nightly and beta), and I receive the correct return. However, something at the level of my work laptop gives a different string when run either from the IDE or manually.

I am waiting for a response from a developer, that covers this area and will provide an update once I have more information. For now, I did a commit with the expected string from my personal laptop.

soncuteanca: I just had a VPN left in background from my manual testing credential management work. All good after I've disabled it. The test passes locally on my end too with the RSA  string.

ben-c-at-moz (approved):
LGTM

PR #115, opened 2024-07-15
Verifying Set and Change Notifications

sarinali:
# Description

Ensures that blocking/allowing notifications has correct behaviour in about:preferences

## Bugzilla bug ID

**Testrail:** https://testrail.stage.mozaws.net/index.php?/cases/view/159150 
**Link:** https://bugzilla.mozilla.org/show_bug.cgi?id=1905173

## Type of change

- [x] New Test

Finishes

# Screenshots / Explanations

N/A

# Comments / Concerns

It is worth noting that the behaviour of the automated test is changed slightly due to the fact that after allowing/blocking the notifications in a session, the popup no longer shows up in the same session. So I separated them into two parameterized situations. 

Another thing to consider is that I was unable to verify a notification popped up, that is one of my blockers. If anyone knows how to fix that, please let me know!


modules/util.py
```diff
@@ -370,6 +370,22 @@ def assert_json_value(self, json_data, jsonpath_expr, expected_value):
             f"Expected {expected_value}, but got {match[0].value}",
         )
 
+    def trim_url_to_last_slash(self, url: str):
```
ben-c-at-moz, line 373: I could be wrong, but I think the general case for what this method does is _actually_ "get the domain of a web resource" rather than "get the parent URL directory". Consider this instead:

```python
from urllib.parse import urlparse, urlunparse

def get_domain_from_url(self, url: str) -> str:
    parsed_url = urlparse(url)
    domain_parsed_url = parsed_url._replace(path="")
    return urlunparse(domain_parsed_url)
```

Up to you, but I think this is going to be more relevant

sarinali: oh yeah, thanks for the suggestion! I just implemented that

modules/data/generic_page.components.json
```diff
@@ -5,6 +5,18 @@
         "groups": []
     },
 
+    "authorize-notiifcations-button": {
```
ben-c-at-moz, line 8: spelling

tests/security_and_privacy/test_notifications_change_set.py
```diff
@@ -0,0 +1,66 @@
...
+    util = Utilities()
+    alert_page = GenericPage(driver, url=NOTIFICATIONS_SITE).open()
+    about_prefs = AboutPrefs(driver, category="privacy")
+
+    alert_page.get_element("authorize-notiifcations-button").click()
+    with driver.context(driver.CONTEXT_CHROME):
+        about_prefs.get_element(button_data).click()
+
+    check_notification_script = """
+    return Notification.permission;
```
ben-c-at-moz, line 39: indent these two lines

ben-c-at-moz (commented):
Lint only, fix that and we're good to go

ben-c-at-moz (approved):
Let's go

PR #116, opened 2024-07-15
rerun failing tests

ben-c-at-moz:
# Description

Allows failing tests to be retried twice before failing out.

***PLEASE NOTE*** New dependency. Please run `pipenv install` before testing.

## Type of change

- [x] QoL, velocity

# How does this resolve / make progress?

Should allow for "mostly stable" tests to pass almost always now. We absolutely will come back to them to try to remove the need for retries.


sarinali (approved):
LGTM, weird failure for one of the tests, I changed the parameterization to `card_indices = [(4, False), (0, True)]` instead of `card_indices = [(3, False), (0, True)]` is this the case for you?

PR #121, opened 2024-07-17
Bc/open link in private

ben-c-at-moz:
# Description

Create test to check that links can be opened in private windows through the context menu.

## Bugzilla bug ID

**Testrail:** https://testrail.stage.mozaws.net/index.php?/cases/view/101662
**Link:** https://bugzilla.mozilla.org/show_bug.cgi?id=1905177

## Type of change

- [x] New Test
- [x] Other Changes -- Fix one of the ContextMenu methods

Test complete

### DIFF
--- a/modules/browser_object_context_menu.py
+++ b/modules/browser_object_context_menu.py
@@ -28,3 +28,11 @@ class ContextMenu(BasePage):
         with self.driver.context(self.driver.CONTEXT_CHROME):
             self.fetch(reference, labels=labels).click()
             return self
+
+    def click_and_hide(
+        self, reference: Union[str, tuple, WebElement], labels=[]
+    ) -> BasePage:
+        with self.driver.context(self.driver.CONTEXT_CHROME):
+            self.fetch(reference, labels=labels).click()
+            self.hide_popup_by_child_node(reference, labels=labels)
+            return self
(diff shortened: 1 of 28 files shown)


modules/browser_object_context_menu.py
```diff
@@ -17,9 +19,20 @@ def get_context_item(self, item: str) -> WebElement:
...
+    ) -> BasePage:
         """
         Clicks the context item.
         """
         with self.driver.context(self.driver.CONTEXT_CHROME):
-            context_element.click()
+            self.fetch(reference, labels=labels).click()
+            return self
+
+    def click_and_hide(
```
sarinali, line 32: much needed!

modules/page_base.py
```diff
@@ -586,15 +586,15 @@ def hide_popup_by_class(self, class_name: str) -> None:
...
+        node = self.fetch(reference, labels=labels)
         script = """var element = arguments[0].parentNode;
                  if (element && element.hidePopup) {
                     element.hidePopup();
                  }"""
-        if chrome:
-            with self.driver.context(self.driver.CONTEXT_CHROME):
-                self.driver.execute_script(script, node)
-        else:
+        with self.driver.context(self.context_id):
```
sarinali, line 597: Is there ever going to be a case where we have to dismiss a context menu in content context? I was just wondering bc of the change here

sarinali (approved):
LGTM, can you just explain that one method to me?

modules/page_base.py
ben-c-at-moz: I don't believe it's even possible to spawn a context menu in the content context.

PR #125, opened 2024-07-18
CI Slack Notifier

sarinali:
# Description

Adds the slack notifier functionality after run of Smoke Tests. Notifies the #desktop-qa-monitoring-notifier channel

## Type of change

- [x] Other Changes: Monitoring

# Screenshots / Explanations

image

# Comments / Concerns

Need to configure this to only run on merge onto main, for now it will run each time someone makes a commit and actions run which is annoying


taskcluster/kinds/run-smoke-tests/kind.yml
```diff
@@ -25,3 +32,63 @@ tasks:
...
+                      "text": {
+                        "type": "mrkdwn",
+                        "text": "*Owner*: ${task.metadata.owner}"
+                      }
+                    },
+                    {
+                      "type": "section",
+                      "text": {
+                        "type": "mrkdwn",
+                        "text": "<https://github.com/mozilla/fx-desktop-qa-automation/commit/${task.metadata.source}|Commit>"
```
ben-c-at-moz, line 71: looks like `task.metadata.source` is the full URL, so you don't need to add the base url to this link.

ben-c-at-moz (changes requested):
Is there a way to get a link to the PR through any of these session variables? If not, then just this one change. If so, I'd love that over the commit link.

taskcluster/kinds/run-smoke-tests/kind.yml
sarinali: Ok, let me look into it

PR #130, opened 2024-07-19
Private Browsing Login Doorhanger

sarinali:
# Description

Ensures that in private browsing the popup for the login that prompts you to save the password does not show up 

## Bugzilla bug ID

**Testrail:** https://testrail.stage.mozaws.net/index.php?/cases/view/101670
**Link:** https://bugzilla.mozilla.org/show_bug.cgi?id=1905183

## Type of change

- [x] New Test
- [x] Other Changes: Sub class of a POM, what do you think of this design? 

Finishes

# Screenshots / Explanations

N/a

# Comments / Concerns

Button indices are hardcoded, on the sample page there are a lot of elements with the same labels/html


ben-c-at-moz (approved):
LGTM

PR #131, opened 2024-07-19
Cookies are not Saved Private Browsing

sarinali:
# Description

Ensures that cookies are not saved in private browsing

## Bugzilla bug ID

**Testrail:** https://testrail.stage.mozaws.net/index.php?/cases/view/101677
**Link:** https://bugzilla.mozilla.org/show_bug.cgi?id=1905186

## Type of change

- [x] New Test

Finishes

# Screenshots / Explanations

N/A

# Comments / Concerns

The `get_all_children` method is slow when there are no children. This causes the test to run ~10 seconds but that is expected behaviour. Perhaps there is a way to fix this?

### DIFF
--- a/tests/security_and_privacy/test_cookies_not_saved_private_browsing.py
+++ b/tests/security_and_privacy/test_cookies_not_saved_private_browsing.py
@@ -1,10 +1,10 @@
-import logging
-from time import sleep
+
 from selenium.webdriver import Firefox
 
+from modules.browser_object import Navigation, PanelUi
 from modules.page_object import AboutPrefs, GenericPage
-from modules.browser_object import Navigation, TabBar, PanelUi
 from modules.util import BrowserActions
+
 
 def test_cookies_not_saved_private_browsing(driver: Firefox):
     """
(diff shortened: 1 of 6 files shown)


tests/security_and_privacy/test_cookies_not_saved_private_browsing.py
```diff
@@ -0,0 +1,47 @@
...
+    # get the cookies
+    about_prefs.get_element("cookies-manage-data").click()
+    iframe = about_prefs.get_iframe()
+    ba.switch_to_iframe_context(iframe)
+
+    # ensure none are listed
+    sitelist = about_prefs.get_element("cookies-manage-data-sitelist")
+    sites = about_prefs.get_all_children(sitelist)
+
+    # note when there are 0 children, this is a bit slow
```
ben-c-at-moz, line 46: This is slow when there are 0 children?

ben-c-at-moz (dismissed):
Looks good!

ben-c-at-moz (approved):
Good work!

PR #132, opened 2024-07-22
Fix to the selectors (using xpath)

sarinali:
Change some of the elements in the manifest to use the xpath instead of hardcoding indices and guessing where it is.


ben-c-at-moz (approved):
LGTM

PR #138, opened 2024-07-24
Cross site trackers, cryptominers and fingerprinters blocked

sarinali:
# Description

Ensures that some trackers are blocked on certain website. Does this by verifying the tracker panel

## Bugzilla bug ID

**Testrail:** https://testrail.stage.mozaws.net/index.php?/cases/view/446393
**Link:** https://bugzilla.mozilla.org/show_bug.cgi?id=1905196

## Type of change

- [x] New Test
- [x] Other Changes: Refactor other tests that use the tracker panel by getting rid of sleep and force refreshing until the shield icon appears to be blocking

Done

# Screenshots / Explanations

N/A

# Comments / Concerns

Hacky as hell, specifically the for loop but since all the children are the same level I can't think of a better way to do this, suggestions open


ben-c-at-moz (changes requested):
Test is not fully passing flake checks for me...

```
E       AssertionError: Not all of the expected items were found in each of the Blocked/Allowed sections.
E       assert (1 == 0)
E        +  where 1 = len({'Fingerprinters'})
```

ben-c-at-moz (approved):
That seems to have done it, good to merge now!

PR #141, opened 2024-07-25
Detected blocked trackers 

sarinali:
# Description

Ensures that on certain websites, the correct type of trackers are listed as "Allowed" and "Blocked"

## Bugzilla bug ID

**Testrail:** https://testrail.stage.mozaws.net/index.php?/cases/view/446392
**Link:** https://bugzilla.mozilla.org/show_bug.cgi?id=1905195

## Type of change

- [x] New Test

Finishes

# Screenshots / Explanations

N/A

# Comments / Concerns

Another test uses the method I created in the BOM in this PR. Since the method is reuseable, in the future `test_trackers_crypto_fingerprint_blocked` will have to be refactored to use it.

### DIFF
--- a/modules/browser_object_tracker_panel.py
+++ b/modules/browser_object_tracker_panel.py
@@ -1,5 +1,5 @@
 import logging
-from typing import List
+from typing import List, Set
 
 from selenium.common.exceptions import TimeoutException
 from selenium.webdriver.remote.webelement import WebElement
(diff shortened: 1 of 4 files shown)


modules/browser_object_tracker_panel.py
```diff
@@ -15,6 +15,58 @@ class TrackerPanel(BasePage):
 
     URL_TEMPLATE = ""
 
+    def verify_allowed_blocked_trackers(
```
ben-c-at-moz, line 18: This is great, but I think it would be more useful if both `allowed` and `blocked` had default empty set values, that way we could use this for all three of: verifying that allowed trackers are allowed, verifying that blocked trackers are blocked, and both at the same time. Currently, this just does both. If there are other methods that do one or the other, they could be replaced by this in the BOM and the tests.

ben-c-at-moz (approved):
One request that I think will help ergonomics around the board, but it's good to go.

ben-c-at-moz (approved):
Good to merge!

PR #142, opened 2024-07-26
Bc/private downloads

ben-c-at-moz:
# Description

* Complete test that private mode downloads are not displayed in normal mode about:downloads
* Add POM: AboutDownloads with methods
* Add BOM: Toolbar with methods
* Add BOM: AboutDownloadsContextMenu with methods
* Ergonomics: improve `get_parent_of`, add `switch_to_alert`, `element_not_visible`, improve reliability of `hide_popup_by_child`.
* Hard waits before starting GHA pipelines

## Bugzilla bug ID

**Testrail:** https://testrail.stage.mozaws.net/index.php?/cases/view/101674
**Link:** https://bugzilla.mozilla.org/show_bug.cgi?id=1905185

## Type of change

- [x] New Test
- [x] New POM
- [x] Ergo, BasePage improve

Completes test

### DIFF
--- a/collect_executables.sh
+++ b/collect_executables.sh
@@ -105,6 +105,7 @@ do
     sleep 0.2
 done
 chmod +x geckodriver
+./geckodriver --version
 
 if [[ $SYSTEM_NAME == "linux" ]]
 then
--- a/modules/page_base.py
+++ b/modules/page_base.py
@@ -8,7 +8,7 @@ from pathlib import Path
 from typing import List, Union
 
 from pypom import Page
-from selenium.common.exceptions import NoSuchElementException, TimeoutException
+from selenium.common.exceptions import NoSuchElementException, TimeoutException, NoAlertPresentException
 from selenium.webdriver import ActionChains, Firefox
 from selenium.webdriver.common.by import By
 from selenium.webdriver.common.keys import Keys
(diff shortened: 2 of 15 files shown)


collect_executables.sh
```diff
@@ -105,6 +105,7 @@ do
     sleep 0.2
 done
 chmod +x geckodriver
+./geckodriver --version
```
sarinali, line 108: Is this to ensure the driver exists during runtime?

modules/page_base.py
```diff
@@ -570,6 +583,19 @@ def switch_to_new_tab(self) -> Page:
...
         return self
 
+    def switch_to_alert(self) -> Page:
+        """Wait for an alert to appear, then switch to it"""
+
+        def alert_present(driver) -> bool:
+            try:
+                driver.switch_to.alert
+                return True
+            except NoAlertPresentException:
```
sarinali, line 593: Is it up to the user using this method to error check and raise the exception (if the alert isnt present)? I was wondering because I don't see any instances of it in the code

modules/page_base.py
```diff
@@ -616,8 +642,7 @@ def hide_popup_by_child_node(
                  if (element && element.hidePopup) {
                     element.hidePopup();
                  }"""
-        with self.driver.context(self.context_id):
-            self.driver.execute_script(script, node)
+        self.driver.execute_script(script, node)
```
sarinali, line 645: I just wanted to know why the driver management (`with self.driver.context(self.context_id)` line) was removed

sarinali (dismissed):
LGTM, can you explain a couple things tho? I'm sure they are thought out design choices

collect_executables.sh
ben-c-at-moz: Yeah, I'm hoping it will be useful for debugging when a GHA run crashes.

modules/page_base.py
ben-c-at-moz: Yeah, I ended up not needing it. I guess it really doesn't strictly fail if there's no alert, which is less than desirable. I think I might just delete this function

modules/page_base.py
ben-c-at-moz: It actually is what made my test pass--looks like that for all cases where we thought we had to send the JS command to the chrome context, the content context would have been fine, but for this instance, the chrome context was not responsive

sarinali (approved):
LGTM

PR #144, opened 2024-07-30
add selenium prefs file

ben-c-at-moz:
Simple PR, add a file that tells us what prefs Selenium changes


selenium_set_prefs.js
```diff
@@ -0,0 +1,174 @@
...
+// DO NOT EDIT THIS FILE.
+//
+// If you make changes to this file while the application is running,
+// the changes will be overwritten when the application exits.
+//
+// To change a preference value, you can either:
+// - modify it via the UI (e.g. via about:config in the browser); or
+// - set it within a user.js file in your profile.
+// - set it using Selenium's webdriver.Firefox Options API
+
```
sarinali, line 12: ok so do you mind explaining how I would use this file? Would I delete prefs when I see fit? Or is this file just for reference of what prefs are set?

ben-c-at-moz: This is just a reference

PR #147, opened 2024-07-31
hy/Test bookmark a website via star button

sv-hyacoub:
# Description

Verify that the Websites can be bookmarked via star-shaped button and check if it's correctly bookmarked

## Bugzilla bug ID

**Testrail: https://mozilla.testrail.io/index.php?/cases/view/2084539**
**Link: https://bugzilla.mozilla.org/show_bug.cgi?id=1910725**

## Type of change

- [x] New Test

N/A

# Screenshots / Explanations

N/A

# Comments / Concerns

N/A


sv-hyacoub:
> Add check to ensure the page was actually bookmarked (no need to check places data base or anything like that). Just revisit the bookmarked page to ensure it shows up as bookmarked. If you're feeling bold, find the bookmark in the sidebar or Bookmarks menu!

Felt bold and made a commit with a new verification.

tests/bookmarks_and_history/test_bookmark_website_via_star_button.py
```diff
@@ -0,0 +1,26 @@
...
+    # instantiate object
+    nav = Navigation(driver)
+    panel = PanelUi(driver)
+
+    # Bookmark the given website and check the bookmark star turned blue
+    driver.get(URL_TO_BOOKMARKED)
+    with driver.context(driver.CONTEXT_CHROME):
+        nav.get_element("star-button").click()
+        nav.get_element("save-bookmark-button").click()
+        assert nav.get_element("blue-star-button").is_displayed()
```
ben-c-at-moz, line 21: I think it would be more stable to do `nav.element_visible("blue-star-button")`, as assert doesn't have a wait built into it

tests/bookmarks_and_history/test_bookmark_website_via_star_button.py
```diff
@@ -0,0 +1,26 @@
...
+
+    # Verify that the bookmark is displayed in bookmarks menu
+    panel.open_bookmarks_menu()
+    with driver.context(driver.CONTEXT_CHROME):
+        assert panel.get_element("mozilla-bookmarked").is_displayed()
```
ben-c-at-moz, line 26: Same as above--I think this will resolve CI failures, try it out! You shouldn't even need to call out the context on line 25 if it works!

ben-c-at-moz (commented):
A couple of things that might resolve the CI issues

tests/bookmarks_and_history/test_bookmark_website_via_star_button.py
sv-hyacoub: I tried nav.element_visible("blue-star-button") with and without assert before it, in both cases the CI failed.

tests/bookmarks_and_history/test_bookmark_website_via_star_button.py
sv-hyacoub: Tried to apply the changes here also, didn't resolve CI failures.

ben-c-at-moz (dismissed):
Please check out my changes before you merge to see how the CI failure was resolved--looks like it had to do with the selector for the Mozilla entry in the bookmarks menu

ben-c-at-moz (approved):

modules/data/panel_ui.components.json
```diff
@@ -143,6 +143,17 @@
...
-    }
+    },
 
+    "panel-ui-bookmarks": {
+        "selectorData": "appMenu-bookmarks-button",
+        "strategy": "id",
+        "groups": []
+    },
+
+    "bookmark-by-title": {
```
ben-c-at-moz, line 154: I think it's possible that the previous version of this was finding another element on the page, and that wasn't a problem for fast computers but it was a problem for the GHA windows machine, which is very slow

sv-hyacoub: I struggled with that selector and I were sure that there is a better way to write that selector.

PR #150, opened 2024-07-31
Reader View Open Close

sarinali:
# Description

Verifying that the reader view can be opened and closed using different methods.

## Bugzilla bug ID

**Testrail:** https://mozilla.testrail.io/index.php?/cases/view/130908
**Link:** https://bugzilla.mozilla.org/show_bug.cgi?id=1910718

## Type of change

- [x] New Test

Finishes

# Screenshots / Explanations

N/A

# Comments / Concerns

Repeated code ngl, is there a way to pass in a function into a function? Also, skipped the one that opens reader view from the top menu. That is inaccessible to selenium


modules/browser_object_reader_view.py
```diff
@@ -7,3 +9,71 @@ class ReaderView(BasePage):
...
+                self.perform_key_combo(Keys.F9)
+        self.wait.until(lambda _: self.driver.page_source != before_page_source)
+        self.wait_for_reader_view_closed()
+        return self
+
+    def wait_for_reader_view_open(self) -> BasePage:
+        """
+        Checks to see if the reader view toolbar is present demonstrating that the reader view is open.
+        """
+        self.wait.until(lambda _: self.element_exists("reader-toolbar"))
```
ben-c-at-moz, line 71: `element_exists` and `element_does_not_exist` use WebDriverWait objects under the hood, we don't need to nest them. That is, this line could just be `self.element_exists("reader-toolbar")`

modules/browser_object_reader_view.py
```diff
@@ -7,3 +9,71 @@ class ReaderView(BasePage):
...
+        return self
+
+    def wait_for_reader_view_closed(self) -> BasePage:
+        """
+        Checks to see if the reader view toolbar is not present, demonstrating that reader view is not open.
+        """
+        self.wait.until(lambda _: self.element_does_not_exist("reader-toolbar"))
```
ben-c-at-moz, line 78: See above

ben-c-at-moz (commented):
Two quick changes, otherwise totally good.

ben-c-at-moz (dismissed):
LGTM!

ben-c-at-moz (approved):
good to merge

PR #153, opened 2024-08-01
get fx beta for mac and win

ben-c-at-moz:
CI change that *should* get the correct version of Fx Beta for Mac and Windows runners -- First few tests in Win still need reruns, not sure how to fix that but it's better than it was before.


sarinali (approved):
after reading through your message and the code several times, I think it makes sense to me. lgtm

PR #157, opened 2024-08-02
run headless and headed tests

ben-c-at-moz:
# Description

Runs tests that need headed mode in their own steps in CI.

* Attempt to fix GHA tests failing early in the session
* Add steps for GHA workflows that run headed tests and save artifacts
* Add steps for TC that do the same thing
* Change mark from `pynput` to `headed` to clarify the purpose
* Remove headless marker in CI
* Remove unstable markers as reasonable - one converted into only unstable in Win GHA
* Fix git hook script to avoid commit failures
* Lint on yaml file

## Bugzilla bug ID

**Link:** https://bugzilla.mozilla.org/show_bug.cgi?id=1892832

## Type of change

- [x] Other Changes - CI

Provides an alternative solution: run pynput tests in headed mode in CI, since all CI machines are VMs at this point.

# Comments / Concerns

Some pynput tests are more stable in *nix than in Win.


ci_pyproject.toml
```diff
@@ -16,7 +16,7 @@ markers = [
 testpaths = [
     "tests"
 ]
-addopts = "-vs --ci --reruns 2 --reruns-delay 1 -m 'not incident and not unstable' --html=artifacts/report.html"
+addopts = "-vs --ci --run-headless --reruns 2 --reruns-delay 1 -m 'not incident and not unstable and not headed' --html=artifacts/report.html"
```
sarinali, line 19: so currently there is no set of flags for the headed runs? or do these flags just add to the base command when we are running typical test runs?

conftest.py
```diff
@@ -70,14 +70,17 @@ def pytest_exception_interact(node, call, report):
...
             logging.error(f"Handling exception for test: {test_name}")
-            logging.error(
-                f"NODE LOGS HERE {node.funcargs}\n THE FAILED TEST: {test_name}"
-            )
-            driver = node.funcargs.get("driver")
-            opt_ci = node.funcargs.get("opt_ci")
-            if driver:
-                log_content(opt_ci, driver, test_name)
-                screenshot_content(driver, opt_ci, test_name)
+            if hasattr(node, "funcargs"):
```
sarinali, line 73: im acc sad that this isnt always defined during runtime ;-; pytest whyyy

sarinali (commented):
Ok so I'm a bit confused right now, I understand that you added some code to ensure that headless runs have the correct arguments and headed tests are not run? I didn't see anything related to that in any of the commands. Do you mind explaining what the plan is for headed runs?

conftest.py
ben-c-at-moz: If the test doesn't get collected properly, the object passed to the hook is `Collector`, which does not have `funcargs`. If the test was collected, it's the normal test `Node` or `Item` or whatever it's called.

ci_pyproject.toml
ben-c-at-moz: All addopts are dynamically added to pytest runs when you call them from the command line. This change removes `headless` as a default option for `ci` and creates a new pyproject file for headed runs, see the CI workflow yamls

ben-c-at-moz:
> Ok so I'm a bit confused right now, I understand that you added some code to ensure that headless runs have the correct arguments and headed tests are not run? I didn't see anything related to that in any of the commands. Do you mind explaining what the plan is for headed runs?

Headed runs are executed in their own step in the CI workflows, after the headless runs

sarinali (approved):
Ok, thanks for clarifying the comments -- lgtm

PR #160, opened 2024-08-05
fix funcargs

ben-c-at-moz:
Not sure what happened, but this is another emergency merge, to _actually_ fix the funcargs error.


ben-c-at-moz:
Emergency merge

PR #162, opened 2024-08-06
Vs/test pdf download

vsangereanMOZ:
# Description

Test that a PDF can be downloaded by clicking on the download button.

## Bugzilla bug ID
1910700[https://bugzilla.mozilla.org/show_bug.cgi?id=1910700]

**Testrail:** C3932
**Link:** https://mozilla.testrail.io/index.php?/cases/view/3932

## Type of change

- [x] New Test
- [ ] New POM

none

# Screenshots / Explanations

N/A

# Comments / Concerns

Could not figure out how to use the get_element instead of find_element


tests/pdf_viewer/test_pdf_download.py
```diff
@@ -0,0 +1,60 @@
...
+
+from modules.browser_object_navigation import Navigation
+from modules.page_object import GenericPdf
+
+
+@pytest.fixture()
+def add_prefs():
+    return []
+
+
```
ben-c-at-moz, line 18: You'll need to add a pytest mark called `headed` for this to run in CI, as you're using pynput. You could also theoretically try setting the `browser.download.manager.showWhenStarting` to `False` to avoid the dialog.

tests/pdf_viewer/test_pdf_download.py
```diff
@@ -0,0 +1,60 @@
...
+def test_pdf_download(driver: Firefox, fillable_pdf_url: str):
```
ben-c-at-moz, line 19: I added a new conftest fixture for you so you don't have to find the download directory inside the test--everyone's going to want to use that code. Check out what I did, and feel free to copy that process in the future when you need to do something that a lot of us are also going to need. Just request `downloads_folder` like you did `fillable_pdf_url` in the function signature.

tests/pdf_viewer/test_pdf_download.py
```diff
@@ -0,0 +1,60 @@
...
+    # Allow time for the download dialog m to appear and pressing enter to download
+    time.sleep(2)
+    keyboard.press(Key.enter)
+    keyboard.release(Key.enter)
+
+    # Allow time for the download to complete
+    time.sleep(2)
+
+    # Determine the platform and set the expected download path
+    this_platform = platform.system()
```
ben-c-at-moz, line 39: This won't be necessary if you use the fixture mentioned above.

tests/pdf_viewer/test_pdf_download.py
```diff
@@ -0,0 +1,60 @@
...
+    saved_pdf_location = ""
+    file_name = "i-9.pdf"
+
+    if this_platform == "Windows":
```
ben-c-at-moz, line 43: Likewise, you can remove all this code if you request the fixture...I basically copied this code and put it in conftest.

ben-c-at-moz (changes requested):
Got very far! There's just a little to fix up so it passes CI

tests/pdf_viewer/test_pdf_download.py
vsangereanMOZ: I searched a lot for this preference in order to avoid the multiple sending of Enter and Tab keys in order to make it work on all OS, but could not find it yet.

tests/pdf_viewer/test_pdf_download.py
vsangereanMOZ: kept file name.

ben-c-at-moz (approved):
good to merge!

PR #164, opened 2024-08-06
CI and Profile Support

ben-c-at-moz:
# Description

* Does not write test failure artifacts unless --ci is set
* Sets Mac and Win Fx to be the correct Beta versions
* Adds a test to grab the Fx version in `meta`
* Adds a fixture in conftest to prep to use profiles in tests (I did check that it works)
* Adds first profile to test with, called `theme_change`.

# Comments / Concerns

Fx and gecko locations are now hardcoded, this should be addressed in a future PR (which should also deprecate `collect_executables.sh`).


ben-c-at-moz:
> Looks great and was able to run a test using the fixture with theme_change profile.
> 
> However, there is still the 'beta' folder with cruft in it that should be removed.

That folder shouldn't be there anymore, it's not in the changeset. Try removing it from your local and repulling (with my branch checked out)

tests/bookmarks_and_history/test_bookmark_website_via_star_button.py
```diff
@@ -13,6 +13,7 @@ def test_bookmark_website_via_star(driver: Firefox):
     # instantiate object
     nav = Navigation(driver)
     panel = PanelUi(driver)
+    # random comment
```
sarinali, line 16: meant to be here?

profiles/theme_change.zip
sarinali, line 1: so im assuming this is unzipped and used right? sorry just want to ask about how it works first

sarinali (approved):
lgtm, can you just answer my question?

tests/bookmarks_and_history/test_bookmark_website_via_star_button.py
ben-c-at-moz: LOOOOL, the change that started it all. No, that should be removed.

profiles/theme_change.zip
ben-c-at-moz: Yep, check out the driver fixture in Conftest.

PR #165, opened 2024-08-07
As/download panel triggered on content disposition attachment

soncuteanca:
# Description

Ensure that the "Always ask" option in Firefox Applications settings triggers the download panel for PDFs with Content-Disposition: attachment.

## Bugzilla bug ID

**Testrail: https://bugzilla.mozilla.org/show_bug.cgi?id=1910705**
**Link: https://mozilla.testrail.io/index.php?/cases/view/936502**

## Type of change

- [x ] New Test

Completed

# Screenshots / Explanations

# Comments / Concerns


vsangereanMOZ (commented):
lgtm

ben-c-at-moz (changes requested):
Looks like we're going to need a path for Linux using pynput--this is a common problem with using auto-GUI libraries in that Linux key paths often differ from Mac and Win...I'll try to find a way around using pynput when I have a chance

ben-c-at-moz (approved):
Changed a selector and its name, but ran into the same problem you were having with the menu items. Instead of pynput, I'm sending keys to elements, and that seems to be working better

ben-c-at-moz (approved):

PR #166, opened 2024-08-07
Remove usages of legacy search bar

Tracy-Walker:
### Description
Due to pref for the legacy search bar being removed in Fx130, we have to remove tests that used that legacy search bar. One test had to be removed altogether as it was strictly testing the legacy search bar.

### Type of change
- [X] Test alterations and removal

Resolves this issue

### Comments / Concerns
We perhaps could have attempted to us D&D from the Customize dialog to place the legacy search bar.  But decided that would be too difficult and flaky.


ben-c-at-moz (approved):
Looks good!

PR #169, opened 2024-08-08
vs/test-pdf-input-numbers

vsangereanMOZ:
# Description

Add a test to check if the numeric values are saved after being input into a field and changing focus

## Bugzilla bug ID

**Testrail:** https://mozilla.testrail.io/index.php?/cases/view/1017528
**Link:** https://bugzilla.mozilla.org/show_bug.cgi?id=1910707

## Type of change

- [ ] New Test
- [ ] New POM

N/A

# Screenshots / Explanations

N/A

# Comments / Concerns

Still a draft


modules/data/generic_pdf.components.json
```diff
@@ -37,5 +37,11 @@
         "selectorData": "download",
         "strategy": "id",
         "groups": []
+    },
+
+    "zipcode-field": {
+        "selectorData": "pdfjs_internal_id_622R",
```
ben-c-at-moz, line 43: I don't necessarily trust PDFjs internal CSS ids, I think a better selector would be `input[name='ZIP Code']` with strategy as `css`.

ben-c-at-moz (changes requested):
Just one change to make

ben-c-at-moz (approved):

PR #173, opened 2024-08-08
Theme Control Panel

sarinali:
# Description

Finishing the type control panels for the sub panel theme

## Bugzilla bug ID

**Testrail:**  https://mozilla.testrail.io/index.php?/cases/view/130919
**Link:** https://bugzilla.mozilla.org/show_bug.cgi?id=1910721

## Type of change

- [x] New Test

Finishes

# Screenshots / Explanations

N/A

# Comments / Concerns

Hm, I remember a discussion between Ben and Tracy about Testrail integration: a file should map to ONE Testrail test, does this mean I should move this in with the other type control panel tests? It is a bit different which is why I grouped it on its own but for the integration's sake it might be better together?

### DIFF
--- a/modules/data/reader_view.components.json
+++ b/modules/data/reader_view.components.json
@@ -101,6 +101,18 @@
         "groups": []
     },
 
+    "toolbar-theme-contrast": {
+        "selectorData": "contrast-button",
+        "strategy": "class",
+        "groups": []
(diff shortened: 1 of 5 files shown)


ben-c-at-moz (approved):
good to merge!

PR #174, opened 2024-08-09
CI and QoL Improvements

ben-c-at-moz:
# Description

Please include a summary of the changes and the related issue. Please also include relevant motivation and context.

* Updates CI and executable collection process
  * Removes hardcoding for GHA bin URLs
  * Moves most collect logic to python
* Factors `delete_files` out of its test and into conftest
* Creates `delete_files_regex_string` to make the previous work
* Factor pynput out of util.py
* Various test marks


collect_executables.sh
```diff
@@ -1,132 +1,14 @@
...
-        ls .
-    else
-        if [[ $SYSTEM_NAME == "macos" ]]
-        then
-            hdiutil attach Firefox*.dmg
-            echo "/Volumes/Firefox/Firefox.app/Contents/MacOS/firefox" > fx_location
-        fi
-    fi
 fi
+curl -o geckodriver.tar.gz -L "$(pipenv run python collect_executables.py -g)"
```
sarinali, line 13: so the flag `-g` is for the geckodriver right?

tests/pdf_viewer/test_pdf_download.py
```diff
@@ -12,9 +12,18 @@ def add_prefs():
     return []
 
 
+@pytest.fixture()
```
sarinali, line 15: I see, so we override the regex in the original conftest to specify what we want to delete right? If that's the case I want to merge this into my MIME types PR to use

sarinali (approved):
Can you just answer my 2 questions? Otherwise looks good!

tests/pdf_viewer/test_pdf_download.py
ben-c-at-moz: yep!

collect_executables.sh
ben-c-at-moz: yep!

PR #177, opened 2024-08-12
GHA Notifier

sarinali:
# Description

Monitoring version made for Windows and Mac

## Type of change

- [x] Other Changes: Monitoring Feature

After each push, PR there will be artifacts uploaded and available for you to view. A slack message will send to notify you that the job is done as well.

# Screenshots / Explanations

image

# Comments / Concerns

https://docs.google.com/document/d/1bpxx9eLmtZ68dc_xE-aXPOepqzsrPPtPLX3GpuN3UMc/edit (some documentation stuff and basic step by step process implementation)


.github/workflows/smoke.yml
```diff
@@ -96,11 +96,52 @@ jobs:
         run: |
           mv ./ci_pyproject_headed.toml ./pyproject.toml;
           pipenv run pytest --fx-executable=/Volumes/Firefox/Firefox.app/Contents/MacOS/firefox -n 4 . || TEST_EXIT_CODE=$?
-          mv -r artifacts/* artifacts-mac/ || true
+          mv artifacts/* artifacts-mac/ || true
```
ben-c-at-moz, line 99: You're right, `-r` is fake. Can we change it to `-n`? I don't want to clobber anything.

tests/pdf_viewer/test_pdf_download.py
```diff
@@ -75,4 +75,3 @@ def test_pdf_download(
     print(
         f"Test passed: The file {file_name} has been downloaded and is present at {saved_pdf_location}."
     )
-    driver.quit()
```
ben-c-at-moz, line 78: Why remove this?

ben-c-at-moz (changes requested):
One change doesn't make sense to me, but other than that, this looks great. (I suspect the change is coming from a linter or something)

tests/pdf_viewer/test_pdf_download.py
sarinali: I think linting deleted this which is super weird, reverting now

.github/workflows/smoke.yml
sarinali: ok! on it

ben-c-at-moz (approved):

PR #191, opened 2024-08-16
Fix Error Catching and Screenshot Functionality

sarinali:
Fixes code written to catch exceptions and failures in tests to screenshot the entire screen and what is captured by the driver. Also ensures that the INTERNAL ERROR does not show up anymore (before an exception was being raised due to failures, now it will simply log and error and move on.)

Note that on Linux CI: wayland machines prevent screenshotting (its a feature...) https://github.com/python-pillow/Pillow/issues/6312 so screenshotting the entire screen is not possible on Linux CI without other libraries. Perhaps this is a come back to feature.

Additional work was done in GHA to ensure that headed tests run if the setup runs correctly and does not depend on the previous unheaded tests to pass anymore.


ben-c-at-moz (dismissed):
Nice work, using pillow was a good idea

ben-c-at-moz (approved):
Looks great, good addition to the tool

PR #192, opened 2024-08-19
Clear all history

sarinali:
# Description

Checks that all time history is deleted and proper ui elements are shown.

## Bugzilla bug ID

**Testrail:** https://mozilla.testrail.io/index.php?/cases/view/172045
**Link:** https://bugzilla.mozilla.org/show_bug.cgi?id=1910748

## Type of change

- [x] New Test

Finishes

# Screenshots / Explanations

N/A

# Comments / Concerns

NA


modules/browser_object_panel_ui.py
```diff
@@ -159,3 +163,27 @@ def open_bookmarks_menu(self) -> BasePage:
...
+    def select_clear_history_option(self, option: str) -> BasePage:
+        """
+        Selects the clear history option, assumes the history panel is open.
+        """
+        with self.driver.context(self.driver.CONTEXT_CHROME):
+            self.get_element("clear-recent-history").click()
+            iframe = self.get_element("iframe")
+            BrowserActions(self.driver).switch_to_iframe_context(iframe)
+
+            with self.driver.context(self.driver.CONTEXT_CONTENT):
```
ben-c-at-moz, line 176: Should we factor Dropdown out of AboutPrefs? This feels counterintuitive...

ben-c-at-moz (commented):
One question about code org, but the test is good.

modules/browser_object_panel_ui.py
sarinali: I think so too, originally was it made for `about:prefs` dropdowns? Where would you prefer this to go? Like its own folder in the `modules`? (eg we make something called like `components` and its `modules/components` or something

ben-c-at-moz: Yeah, that would work! I'm kinda sad we haven't done more of that for this project, honestly.

ben-c-at-moz (approved):
This is great, hopefully we'll use the components folder more often...

PR #203, opened 2024-08-30
Initial entry for SELECTOR_INFO.md

Tracy-Walker:
# Description
Create SELECTOR_INFO.md with first entry as example for others to follow

## Bugzilla bug ID
**Link: https://bugzilla.mozilla.org/show_bug.cgi?id=1915955**

## Type of change
- [X] Documentation

Initiates the documentation of selectors

# Comments / Concerns
Other bugs are on file for each .json file.

### DIFF
--- a/SELECTOR_INFO.md
+++ b/SELECTOR_INFO.md
@@ -1,0 +1,12 @@
+### Selector Information
+This document describes the many DOM selectors used throughout the project, in the format:
+
+**Selector name: (declaration)
+Description: (of the selector, including purpose/functionality)
+Location: (of the selector in content page or Fx UI)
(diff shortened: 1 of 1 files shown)


SELECTOR_INFO.md
```diff
@@ -0,0 +1,12 @@
+### Selector Information
+This document describes the many DOM selectors used throughout the project, in the format:
+
+**Selector name: (declaration)
```
ben-c-at-moz, line 4: What's the reason for the `**` here?

ben-c-at-moz (approved):
This looks like a good start, not sure about one piece of markdown, but looks like the right strategy

PR #229, opened 2024-09-23
Anca/ about:logins navigation from hamburger menu

soncuteanca:
### Description

Verify that clicking the Password option in Hamburger Menu opens about:logins page in a new tab

### Bugzilla bug ID

**Testrail: https://mozilla.testrail.io/index.php?/cases/view/2241082**
**Link: https://bugzilla.mozilla.org/show_bug.cgi?id=1920177**
 
### Type of change

- [ x] New Test

Completed

### Comments / Concerns


tests/password manager/conftest.py
ben-c-at-moz, line 1: Please use an underscore _ instead of a space for the folder name.

Also, if you'd like, you can move the contents of the `tests/credential` folder over there as well, merging the conftest files. You don't have to, though.

tests/password manager/test_about_login_navigation_from_hamburger_menu.py
```diff
@@ -0,0 +1,17 @@
+from selenium.webdriver import Firefox
+
+from modules.browser_object import PanelUi, TabBar
+
```
ben-c-at-moz, line 4: Please add a fixture called `test_case` which returns the test case number (without the C) as a string.

ben-c-at-moz (changes requested):
Test looks good and works well, but there are a couple of minor but important superficial things to fix here.

ben-c-at-moz (approved):
Thanks for fixing this up! Ready to merge

PR #236, opened 2024-09-27
Fix test misusing context menu item clicking

Tracy-Walker:
### Description
I discovered some tests were improperly using `context_click_item()` instead of `click_and_hide_menu()`. The latter clicks the item and ensures the context menu is closed. We should favor its use, unless it is needed by the test to have the context menu remain open.   `click_and_hide_menu()` ensures menus are closed and not blocking next steps.  This PR refactors tests that didn't need to keep the menu open after the menu option was clicked.

#### Bugzilla bug ID
none filed

#### Type of change
- [X] test case stabilization


ben-c-at-moz (approved):
LGTM!

PR #237, opened 2024-10-01
Ben/testrail full

ben-c-at-moz:
### Description

Creates TestRail reporting system that will run given the the environment contains `TESTRAIL_REPORT=1`:

* Finds relevant TR milestone given Fx version under test
* Finds relevant submile based on channel (currently Beta and Alpha are supported, Release will be done in future)
* Finds or creates relevant test plan
* Finds or creates relevant config based on machine architecture
* Finds or creates relevant test plan entry based on config
* Finds or creates relevant test run
* Updates results based on pass, fail, or skip. (Deselected tests are not considered skipped.)

Proof of concept inside the PR:

* Updates the autofill tests to be `xfail`, so they are listed as pass or fail

### Bugzilla bug ID

**Testrail:** Example: https://mozilla.testrail.io/index.php?/plans/view/91747
**Link:** https://bugzilla.mozilla.org/show_bug.cgi?id=1921820

### Type of change

- [x] Structural, reporting

Completes the logic ticket. The wiring ticket is future work, and is not covered by this PR. Currently, nothing happens automatically. This is not the PR for that.

In order to successfully run the code, one needs `TESTRAIL_REPORT`, `TESTRAIL_USERNAME`, `TESTRAIL_API_KEY`, and `TESTRAIL_BASE_URL` set in the environment.


jjSDET (changes requested):
i recently had to fix my logic in what i think is the same use case for the method i had written `does_milestone_exist` because if the response has more than 250 milestones then you need to handle pagination.

fortunately there are new filters available for the `_get_milestones` api that you can pass in.

here's what i'm using now for the same task in taskcluster for `notify_testrail_release` in `mozilla-central`

```python
def does_milestone_exist(self, project_id, milestone_name):
        """
        Check if a milestone with a specific name exists among open milestones.

        Args:
            project_id (int): ID of the project.
            milestone_name (str): Name of the milestone to search for.

        Returns:
            bool: True if the milestone exists, False otherwise.
        """
        if not all([project_id, milestone_name]):
            raise ValueError("Project ID and milestone name must be provided.")

        # Fetch open milestones (is_completed=False)
        response = self._get_milestones(project_id, is_completed=False)
        milestones = response.get("milestones", [])
        # Check for the milestone name
        return any(milestone["name"] == milestone_name for milestone in milestones)
```
        
notice the update to `_get_milestones` i had to make in order to take any number of keyword args `kwargs`, one of which will look for all currently opened milestones (which hopefully is only the most recent milestones currently being worked on, which should then guarantee you're looking at the last 'page' and then check each of those milestones.
        
just mentioning all that because my script broke for that exact reason when the testrail project ended up having over 1k milestones.

modules/testrail.py
```diff
@@ -0,0 +1,289 @@
...
+            f"get_cases/{testrail_project_id}&suite_id={testrail_test_suite_id}"
+        )
+
+    def _update_test_run_results(self, testrail_run_id, data):
+        return self.client.send_post(f"add_results_for_cases/{testrail_run_id}", data)
+
+    def _get_milestones(self, testrail_project_id):
+        milestone_obj = self.client.send_get(f"get_milestones/{testrail_project_id}")
+        return milestone_obj.get("milestones")
+
```
jjSDET, line 264: i regret my decision to return the `milestones` array in this return here given the new filters added with the testrail version bump, which is why i no longer do that in my implementation and instead opt to specify the `milestones` array as a composition of the response when i need it instead, but do whatever works best for your use case.

in order for my recommendation to work for checking if the milestone exists in the comment above, you'll have to update your method here to something like this:

```python
def _get_milestones(self, project_id, **filters):
        """
        Retrieves milestones for a given project with optional filters.

        Args:
            project_id (int): The ID of the project.
            **filters: Arbitrary keyword arguments representing API filters.

        Available Filters:
            is_completed (bool or int):
                - Set to True or 1 to return completed milestones only.
                - Set to False or 0 to return open (active/upcoming) milestones only.

            is_started (bool or int):
                - Set to True or 1 to return started milestones only.
                - Set to False or 0 to return upcoming milestones only.

            limit (int):
                - The number of milestones the response should return.
                - The default response size is 250 milestones.

            offset (int):
                - Where to start counting the milestones from (the offset).
                - Used for pagination.

        Returns:
            dict: The API response containing milestones.
        """

        if not project_id:
            raise ValueError("Project ID must be provided.")

        # Base endpoint
        endpoint = f"get_milestones/{project_id}"

        # Process filters
        if filters:
            # Convert boolean values to integers (API expects 1 or 0)
            for key in ["is_completed", "is_started"]:
                if key in filters and isinstance(filters[key], bool):
                    filters[key] = int(filters[key])

            # Build query parameters
            query_params = "&".join(f"{key}={value}" for key, value in filters.items())
            endpoint = f"{endpoint}&{query_params}"

        # Make API call
        return self.client.send_get(endpoint)
```

i know it's a lot more code, but it's also more flexible now. 

the only thing i'd change about it looking back on it, is to add type hints so those doc strings aren't needed any more.

jjSDET (changes requested):

jjSDET (changes requested):
make sure to add the MPL in your file here. 

i'd also recommend some comments at the top summarizing all your logic here cause i'm a bit lost on ALL the things you're trying to do here. i wonder if it's too much all in one place?

```python
#!/usr/bin/env python3

# This Source Code Form is subject to the terms of the Mozilla Public
# License, v. 2.0. If a copy of the MPL was not distributed with this
# file, You can obtain one at http://mozilla.org/MPL/2.0/.
# TestRail API Wrapper Module

# This module provides methods to interact with the TestRail API.
# API Version: v2
# TestRail Version: 8.0.5 https://mozilla.testrail.io

# For detailed information on the TestRail API, please refer to the official documentation:
# https://support.testrail.com/hc/en-us/sections/7077185274644-API-reference
"""
This module provides a TestRail class for interfacing with the TestRail API, enabling the creation and management of test milestones, test runs, and updating test cases. It facilitates automation and integration of TestRail functionalities into testing workflows, particularly for projects requiring automated test management and reporting.

The TestRail class encapsulates methods to interact with TestRail's API, including creating milestones and test runs, updating test cases, and checking the existence of milestones. It also features a method to retry API calls, enhancing the robustness of network interactions.

Key Components:
- TestRail Class: A class providing methods for interacting with TestRail's API.
  - create_milestone: Create a new milestone in a TestRail project.
  - create_milestone_and_test_runs: Create a milestone and associated test runs for multiple devices in a project.
  - create_test_run: Create a test run within a TestRail project.
  - does_milestone_exist: Check if a milestone already exists in a TestRail project.
  - update_test_cases_to_passed: Update the status of test cases to 'passed' in a test run.
- Private Methods: Utility methods for internal use to fetch test cases, update test run results, and retrieve milestones.
- Retry Mechanism: A method to retry API calls with a specified number of attempts and delay, improving reliability in case of intermittent network issues.

Usage:
This module is intended to be used as part of a larger automated testing system, where integration with TestRail is required for test management and reporting.

"""

import os
import sys
import time
```

modules/testrail.py
ben-c-at-moz: OK, I just went ahead and pasted this method in--it doesn't really change my workflow, but if it's more future-proof I'm for it.

ben-c-at-moz:
> make sure to add the MPL in your file here.
> 
> i'd also recommend some comments at the top summarizing all your logic here cause i'm a bit lost on ALL the things you're trying to do here. i wonder if it's too much all in one place?
> 
> ```python
> #!/usr/bin/env python3
> 
> # This Source Code Form is subject to the terms of the Mozilla Public
> # License, v. 2.0. If a copy of the MPL was not distributed with this
> # file, You can obtain one at http://mozilla.org/MPL/2.0/.
> # TestRail API Wrapper Module
> 
> # This module provides methods to interact with the TestRail API.
> # API Version: v2
> # TestRail Version: 8.0.5 https://mozilla.testrail.io
> 
> # For detailed information on the TestRail API, please refer to the official documentation:
> # https://support.testrail.com/hc/en-us/sections/7077185274644-API-reference
> """
> This module provides a TestRail class for interfacing with the TestRail API, enabling the creation and management of test milestones, test runs, and updating test cases. It facilitates automation and integration of TestRail functionalities into testing workflows, particularly for projects requiring automated test management and reporting.
> 
> The TestRail class encapsulates methods to interact with TestRail's API, including creating milestones and test runs, updating test cases, and checking the existence of milestones. It also features a method to retry API calls, enhancing the robustness of network interactions.
> 
> Key Components:
> - TestRail Class: A class providing methods for interacting with TestRail's API.
>   - create_milestone: Create a new milestone in a TestRail project.
>   - create_milestone_and_test_runs: Create a milestone and associated test runs for multiple devices in a project.
>   - create_test_run: Create a test run within a TestRail project.
>   - does_milestone_exist: Check if a milestone already exists in a TestRail project.
>   - update_test_cases_to_passed: Update the status of test cases to 'passed' in a test run.
> - Private Methods: Utility methods for internal use to fetch test cases, update test run results, and retrieve milestones.
> - Retry Mechanism: A method to retry API calls with a specified number of attempts and delay, improving reliability in case of intermittent network issues.
> 
> Usage:
> This module is intended to be used as part of a larger automated testing system, where integration with TestRail is required for test management and reporting.
> 
> """
> 
> import os
> import sys
> import time
> ```

Did that, but MPL2 is indicated in my license.md, so I thought I was fine...nbd for me though.

jjSDET (dismissed):
lgtm

i'll only note that pagination may be required in the future for any api calls returning a list where the actual number of elements that exist is greater than the `limit`, which defaults to 250.

this may apply to `plans`, `test_cases`, etc.

PR #241, opened 2024-10-02
Anca/update login via update password doorhanger

soncuteanca:
### Description

- Verify that Firefox successfully updates login credentials via the update password doorhanger and
    autofill them on subsequent access

### Bugzilla bug ID

**Testrail: https://mozilla.testrail.io/index.php?/cases/view/2243013**
**Link: https://bugzilla.mozilla.org/show_bug.cgi?id=1920202**

### Type of change

- [x ] New Test

- In progress

- N/A

### Comments / Concerns

- In this test, I can only verify the length of the password field's value, not the actual password itself. this limitation exists because the password field content is masked (for security reasons) and cannot be directly accessed. Therefore, to validate the presence of a password, I assert that the length of the masked value matches the expected length.

### DIFF
--- a/modules/data/login_autofill.components.json
+++ b/modules/data/login_autofill.components.json
@@ -34,19 +34,25 @@
     "username-login-field": {
         "selectorData": "/html/body/div[1]/form[2]/input[1]",
         "strategy": "xpath",
-        "groups": []
+        "groups": [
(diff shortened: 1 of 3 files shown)


ben-c-at-moz (approved):
Good to merge

PR #245, opened 2024-10-06
Add test case for C2241121 - Edit Mode - Change Everything

chickenleaf:
### Description

This PR adds a new test for **C2241121 - Edit Mode - Change Everything**. The test automates the scenario where a login entry is created and then edited using the shadow DOM elements. It interacts with elements such as the add login button, edit button, username, password fields, and save button.

### Bugzilla bug ID

**Testrail:**   C2241121
**Link:**  https://bugzilla.mozilla.org/show_bug.cgi?id=1920185

### Type of change

- [ ] New Test

This test automates the verification of the edit mode functionality in the password manager. It ensures that a login entry can be edited, including changes to the username and password, and successfully saved.

No screenshots are included, but the test verifies that the edit and save functionality work correctly by interacting with shadow DOM elements for the login form.

### Comments / Concerns

Shadow DOM interactions might require updates. I attempted to use the PomUtils class, but it was not compatible with interacting through shadow DOM in this scenario.

### DIFF
--- a/tests/password_manager/test_edit_login_change_everything.py
+++ b/tests/password_manager/test_edit_login_change_everything.py
@@ -1,0 +1,72 @@
+import pytest
+from selenium.webdriver import Firefox
+
+from modules.page_object_about_pages import AboutLogins
+
+@pytest.fixture()
+def test_case():
+    return "2241121"
+
+def test_edit_mode_change_everything(driver: Firefox):
+    """
+    C2241121 - Edit Mode - Change Everything
+    """
+    # Initialize page object and open AboutLogins page
+    about_logins = AboutLogins(driver).open()
+
+    # Click add login button
+    about_logins.click_add_login_button()
+
+    # Create initial test login data and add new login
+    initial_login = {
+        "origin": "test.mozilla.org",
+        "username": "test_user",
+        "password": "test_password123"
+    }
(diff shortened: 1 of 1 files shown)


tests/password_manager/test_edit_login_change_everything.py
```diff
@@ -0,0 +1,72 @@
...
+        "password": "test_password123"
+    }
+    about_logins.create_new_login(initial_login)
+
+    # Click Edit button through shadow DOM
+    assert driver.execute_script("""
+        const editButton = document.querySelector('login-item').shadowRoot.querySelector('edit-button').shadowRoot.querySelector('button.ghost-button');
+        if (editButton) { editButton.click(); return true; }
+        return false;
+    """), "Edit button not found."
```
b4handjr, line 33: This style of finding elements is nothing like we do for this project, please reference other tests and see how it is done and update yours accordingly.

b4handjr (changes requested):
Thanks for the PR but I think this will need a major update to follow the style of the current tests we have.

ben-c-at-moz:
Hi, thanks for the attempt but at this time we need to focus on completing this test before a due date and there doesn't seem to be a timely response to original change requests, so we're going to close this PR. There will be more opportunities in this and other repos in the future. Thank you and have a nice day.

PR #260, opened 2024-10-10
Pat/added test for deny geolocation permissions

whlpatricia:
### Description
Added test to Verify that denying geolocation permissions prevents website from accessing geolocation

### Bugzilla bug ID

**Testrail:** https://mozilla.testrail.io/index.php?/cases/view/122613
**Link:** https://bugzilla.mozilla.org/show_bug.cgi?id=1920215

### Type of change

- [x] New Test

Completed

NA

### Comments / Concerns

NA

### DIFF
--- a/modules/data/navigation.components.json
+++ b/modules/data/navigation.components.json
@@ -415,6 +415,12 @@
         "groups": []
     },
 
+    "popup-notification-block": {
+        "selectorData": "popup-notification-secondary-button",
+        "strategy": "class",
+        "groups": []
+    },
+
     "toolbar-history-recently-closed-tabs": {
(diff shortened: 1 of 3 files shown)


ben-c-at-moz (dismissed):
So this is fine to merge, BUT I would recommend changing `popup-notification-block` to `popup-notification-secondary-button` because the secondary slot is used by different popup notis for different operations.

whlpatricia:
> So this is fine to merge, BUT I would recommend changing `popup-notification-block` to `popup-notification-secondary-button` because the secondary slot is used by different popup notis for different operations.

Makes sense, I will change that

ben-c-at-moz (approved):

PR #263, opened 2024-10-11
Create test automation for test case - Edit Mode (about:logins)

Temidayo32:
### Description

This new test case ensures updates to saved login credentials are successful. For this test case, the focus is on ensuring edits to passwords are successful.

### Bugzilla bug ID

**Testrail:** https://mozilla.testrail.io/index.php?/cases/view/2241121
**Link:** https://bugzilla.mozilla.org/show_bug.cgi?id=1920185

### Type of change

- [x] New Test
- [ ] New POM

This awaits review from project maintainers.

### DIFF
--- a/tests/password_manager/test_about_logins_edit_password.py
+++ b/tests/password_manager/test_about_logins_edit_password.py
@@ -1,0 +1,65 @@
+import pytest
+
+from modules.page_object import AboutLogins
+from modules.util import BrowserActions
+from faker import Faker
+
+
+@pytest.fixture()
+def test_case():
+    return "2241083"
+
+@pytest.fixture()
+def temp_selectors():
+    return {
+        "edit-button": {
+            "selectorData": "edit-buttlogin-item-typeon",
+            "strategy": "tag",
+            "groups": [],
+        },
+        "show-password-button": {
+            "selectorData": ".detail-row .reveal-password-wrapper input.reveal-password-checkbox",
+            "strategy": "css",
+            "shadowParent": "login-item",
+            "groups": []
+        },
+        "not-allowed": {"selectorData": "error", "groups": []},
+    }
+
+
+def test_about_logins_edit_password(driver_and_saved_logins, faker: Faker):
+    """
+    C2241083 - Verifies editing the password of an existing login entry in the about:logins page.
+    """
+    (driver, usernames, logins) = driver_and_saved_logins
+    ba = BrowserActions(driver)
+    about_logins = AboutLogins(driver).open()
+
+    # Step 1: fliter saved logins by username (expect 1 result)
+    ba.clear_and_fill(about_logins.get_element("login-filter-input"), usernames[-1])
+
+    # Step 2: Click the result
+    login_results = about_logins.get_elements("login-list-item")
+    first_login_result = login_results[0]
+    first_login_result.click()
+
+    # Step 3: Click the 'Edit' button
+    about_logins.get_element("edit-button").click()
+
+    new_username = faker.user_name()
+    new_password = faker.password(length=15)
+
+    # Fill in the new username and password
+    ba.clear_and_fill(about_logins.get_element("about-logins-page-username-field"), new_username, press_enter=False)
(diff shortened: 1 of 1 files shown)


tests/password_manager/test_about_logins_edit_password.py
```diff
@@ -0,0 +1,54 @@
...
+    """
+    C2241083 - Verifies editing the password of an existing login entry in the about:logins page.
+    """
+    (driver, usernames, logins) = driver_and_saved_logins
+    ba = BrowserActions(driver)
+    about_logins = AboutLogins(driver).open()
+    ba.clear_and_fill(about_logins.get_element("login-filter-input"), usernames[-1])
+    login_results = about_logins.get_elements("login-list-item")
+    first_login = login_results[0]
+    driver.execute_script(
```
ben-c-at-moz, line 23: We generally don't want arbitrary `execute_script` calls in the tests, it makes them difficult to maintain in the future. Please convert selectors to entries in `modules/data/about_logins.components.json`, add methods to the relevant Page Object in `modules/page_object_about_pages.py` if necessary, and execute commands from that object or the BasePage object in order to complete the test. Thanks!

tests/password_manager/test_about_logins_edit_password.py
```diff
@@ -0,0 +1,54 @@
...
+    displayed_password = about_logins.get_element("about-logins-page-password-field").get_attribute("value")
+
+    shadow_host = about_logins.get_element("login-items")
+
+    driver.execute_script("""
+        const shadowRoot = arguments[0].shadowRoot;
+        const saveChangesButton = shadowRoot.querySelector('.save-changes-button');
+        saveChangesButton.click();
+    """, shadow_host)
+    assert displayed_password == new_password
```
ben-c-at-moz, line 53: In this instance, `assert` is likely to be unstable, as it won't wait for the values to match. You can try `about_logins.wait.until()` with a lambda inside if you want to check something that may not update instantly.

ben-c-at-moz (changes requested):
Please update the test per comments -- all the `execute_script` function calls need to be replaced, and the last assert dealt with, then resubmit. Thanks for contributing!

tests/password_manager/test_about_logins_edit_password.py
Temidayo32: Alright, thanks for the review. I will do that promptly.

Temidayo32:
> Please update the test per comments -- all the `execute_script` function calls need to be replaced, and the last assert dealt with, then resubmit. Thanks for contributing!

Hi, so I tried replacing all of the `execute_script`, but replacing some posed quite a difficulty. This is especially true for shadow DOM elements that are not supported natively by selenium. I created a method on the page object, which relies on the `execute_script` method to trigger a `click` event on a shadow element that continued to remain `elementNotInteractable` when using Selenium `click` method.

ben-c-at-moz:
> > Please update the test per comments -- all the `execute_script` function calls need to be replaced, and the last assert dealt with, then resubmit. Thanks for contributing!
> 
> Hi, so I tried replacing all of the `execute_script`, but replacing some posed quite a difficulty. This is especially true for shadow DOM elements that are not supported natively by selenium. I created a method on the page object, which relies on the `execute_script` method to trigger a `click` event on a shadow element that continued to remain `elementNotInteractable` when using Selenium `click` method.

You should be able to add selectors that exist in shadow DOM much in the same way that the `create-login-button` is selected in `modules/data/about_logins.components.json` -- by using a `shadowParent`. The BasePage method `get_element` (and other functions that depend on it, like `click_on`) will find an element even in a shadow DOM (or iframe).

Try again and re-request review when you're done. Thanks so much!

ben-c-at-moz:
Hi, looks like we are running up against our deadline for this test, so we've gone ahead and had someone else complete it. Your effort was really good, just needed to remove that last script execution. That said, there are more opportunities to contribute to Desktop Test Engineering efforts soon, through the FoxPuppet repo. Thanks for all your work, but we'll be closing this PR now.

Temidayo32:
Actually, I am still working on it.

PR #265, opened 2024-10-14
Anca/ Trigger primary password prompt on about:logins access via hamburger menu

soncuteanca:
### Description

- Verify that the primary password prompt is triggered on accessing about:logins through Hamburger Menu.

### Bugzilla bug ID

**Testrail: https://mozilla.testrail.io/index.php?/cases/view/2245199**
**Link: https://bugzilla.mozilla.org/show_bug.cgi?id=1920207**

### Type of change

- [x ] New Test

- Completed

- N/A

### Comments / Concerns

- Note that this test requires a restart, so I had to work around it. Without restarting the browser after the primary password was set, Firefox does not trigger the primary password prompt when simply opening about:logins because no encrypted data is accessed initially. The prompt is triggered only when sensitive data, is accessed, requiring decryption. Once decrypted data has been accessed (e.g., showing a saved password), Firefox recognizes the need to protect further access to that data. As a result, subsequent navigation to about:logins will immediately trigger the primary password prompt to prevent unauthorized access to the decrypted source.

### DIFF
--- a/SELECTOR_INFO.md
+++ b/SELECTOR_INFO.md
@@ -947,13 +947,6 @@ Description: Shadow parent of the OK and cancel button
 Location: about:preferences#privacy Primary Password popup
 Path to .json: modules/data/about_prefs.components.json
 ```
-```
-Selector Name: set-primary-password-prompt-message
-Selector Data: "commonDialogWindow"
-Description: Prompt message after setting primary password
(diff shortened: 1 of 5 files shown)


ben-c-at-moz (approved):
Good work, this was a tough one!

PR #279, opened 2024-10-21
fix: make cron schedule a multiple of 15 minutes

bhearsum:
I'm not sure why, but this is required per https://github.com/mozilla-releng/fxci-config/blob/410dfc7d9bd24da65800a914fc6faeb5ab1571cf/build-decision/src/build_decision/cron/schema.yml#L65-L69


bhearsum:
oops, already done in #278

PR #282, opened 2024-10-22
As/Geolocation via w3c api

soncuteanca:
### Description

- Verify that geolocation is successfully shared when the user allows permission via the W3C Geolocation API
- Verify that geolocation is not shared when the user blocks permission via the W3C Geolocation API

### Bugzilla bug ID

**Testrail: https://mozilla.testrail.io/index.php?/cases/view/15186**
**Link: https://bugzilla.mozilla.org/show_bug.cgi?id=1925550**

### Type of change

- [x ] New Test

- Completed

- N/A

### Comments / Concerns

- I split the test into two parts to account for the slight differences in behavior between allowing and blocking geolocation (note that the allow permission is temporarily retained for one hour or until the tab is closed, which is not the case for the block setting). Since the steps are repetitive, this also helps isolate the test cases and enhances overall stability.


tests/geolocation/test_geolocation_shared_via_w3c_api.py
```diff
@@ -0,0 +1,161 @@
...
+    """
+    try:
+        driver.switch_to.window(driver.window_handles[-1])
+        web_page.find_element(By.ID, "accept-choices").click()
+    except NoSuchElementException:
+        # If the cookie banner is not found, continue with the test
+        pass
+
+
+def handle_geolocation_prompt(nav, button_type="primary"):
```
ben-c-at-moz, line 43: Could you move this to the navigation browser object? I think this could be useful in other tests in the future...

tests/geolocation/test_geolocation_shared_via_w3c_api.py
```diff
@@ -0,0 +1,161 @@
...
+        By.CSS_SELECTOR, "div[aria-label='You are here!'][role='img']"
+    )
+
+    if location_marker:  # Check if the marker exists
+        return location_marker[0].is_displayed()
+    else:
+        return False
+
+
+def open_web_page_in_new_tab(web_page, driver, tabs, expected_num_tabs):
```
ben-c-at-moz, line 76: could you move this to page_base.py? I think this would be useful for multiple tests...

ben-c-at-moz (dismissed):
Good code, you can merge if you want to, but I do think two things should be moved to other files for re-use.

tests/geolocation/test_geolocation_shared_via_w3c_api.py
soncuteanca: I've moved it for now  to browser_object_tabbar instead. I might be mistaken, but I believe it's more appropriate there.

PR #293, opened 2024-10-23
Ben/http site

ben-c-at-moz:
### Description

Creates test to ensure that HTTP sites can still be visited, and that users are informed that these connections are insecure.

### Bugzilla bug ID

**Testrail:** https://mozilla.testrail.io/index.php?/cases/view/2300294
**Link:** https://bugzilla.mozilla.org/show_bug.cgi?id=1925595

### Type of change

- [x] New Test
- [ ] New POM
- [x] Other Changes - Renames the scroll test suite folder because hyphens and python are not friends

Completes

### Comments / Concerns

Test site is a bit slow


whlpatricia (approved):
lgtm!

PR #297, opened 2024-10-24
Ben/print preview

ben-c-at-moz:
### Description

Complete test for Print Preview modal.

### Bugzilla bug ID

**Testrail:** https://mozilla.testrail.io/index.php?/cases/view/965139
**Link:** https://bugzilla.mozilla.org/show_bug.cgi?id=1925545

### Type of change

- [x] New Test
- [x] New POM
- [x] Other Changes - Update functions in BasePage

Completes test

### Comments / Concerns

We don't do a particularly thorough job of ensuring the correct modal is there, so if someone created, for example, a second modal with the same XPath, we wouldn't know it. It's fine though.

### DIFF
--- a/SELECTOR_INFO.md
+++ b/SELECTOR_INFO.md
@@ -2872,6 +2872,21 @@ Description: Password button
 Location: Hamburger Menu
 Path to .json: modules/data/panel_ui.components.json
 ```
+```
+Selector name: print-option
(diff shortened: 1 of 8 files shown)


vsangereanMOZ (approved):
Read through it and it looks good to me.

PR #300, opened 2024-10-24
Ben/testrail taskcluster l3

ben-c-at-moz:
### Description

We've been upgraded to a level 3 repo and need to have our own TC secrets. I made those TC secrets and now we need to use them in our cron work.

### Type of change

Allows us to see if there are any other blockers to TC cron.


whlpatricia (approved):
lgtm!

PR #303, opened 2024-10-25
Pat/test pasting image and text

whlpatricia:
### Description

Verify that an user can "Paste" image/text in demopage using right click to copy and ctrl+v to paste

### Bugzilla bug ID

**Testrail:** https://mozilla.testrail.io/index.php?/cases/view/464474
**Link:** https://bugzilla.mozilla.org/show_bug.cgi?id=1925568

### Type of change

- [x] New Test

Completed

NA

### Comments / Concerns

NA

### DIFF
--- a/modules/page_base.py
+++ b/modules/page_base.py
@@ -541,16 +541,11 @@ class BasePage(Page):
         with self.driver.context(self.context_id):
             el = self.fetch(reference, labels)
             self.scroll_to_element(el)
-            self.actions.click(el).perform()
-            if sys_platform == "Darwin":
-                self.actions.key_down(Keys.COMMAND)
-                self.actions.send_keys("v")
-                self.actions.key_up(Keys.COMMAND)
-            else:
-                self.actions.key_down(Keys.CONTROL)
-                self.actions.send_keys("v")
-                self.actions.key_up(Keys.CONTROL)
-            self.actions.perform()
+            self.click_on(el)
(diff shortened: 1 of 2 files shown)


modules/page_base.py
```diff
@@ -533,6 +533,51 @@ def context_click(
...
         return self
+    
+    def paste_to_element(
+        self, sys_platform, reference: Union[str, tuple, WebElement], labels=[]
+    ) -> Page:
+        """ Paste the copied item into the element"""
+        with self.driver.context(self.context_id):
+            el = self.fetch(reference, labels)
+            self.scroll_to_element(el)
+            self.actions.click(el).perform()
```
ben-c-at-moz, line 544: Do you need to use ActionChains here? I think `el.click()` or `self.click_on(el)` would be fine

modules/page_base.py
```diff
@@ -533,6 +533,51 @@ def context_click(
...
+            if sys_platform == "Darwin":
```
ben-c-at-moz, line 545: This is fine, but you can factor out the key with

```Python
mod_key = Keys.COMMAND if sys_platform == "Darwin" else Keys.CONTROL
self.actions.key_down(mod_key)
...
```

ben-c-at-moz (approved):
Fine as is, but there are some potential organizational changes that could tighten it up.

modules/page_base.py
whlpatricia: changing to this as well

ben-c-at-moz (approved):

PR #304, opened 2024-10-25
GHA TR reporting: literally fix everything in the world with indentation

ben-c-at-moz:
Things are failing in Windows GHA, possible because of the ~*°=magic=~¨´ of semantic whitespace. I think we can fix this.


whlpatricia (approved):
looks good!

PR #305, opened 2024-10-28
only report when report is available

ben-c-at-moz:
Revert the "indent solution" from previous attempts, then only report to testrail if you have the report. This should fix the Win headless problem.


whlpatricia (approved):
looks good!

whlpatricia (approved):
LGTM

PR #309, opened 2024-10-30
Ben/py crypto for svc accts

ben-c-at-moz:
### Description

We need a solution to encrypt test accounts for services like GMail, as they are needed for certain tests. This leverages Python `cryptography` library to accomplish this with a simple key that can live in normal secret management.

### Type of change

Quality of Life, tool script, fixture, module

Explained in README

### Comments / Concerns

We might want to rework this in future to only encrypt the secrets and leave the JSON structure, like sops does.


whlpatricia (approved):
amazing!

whlpatricia (approved):
Looks good!

PR #311, opened 2024-11-01
Ben/check tr for used

ben-c-at-moz:
### Description

We check for version when we run TestRail report checks against new betas. For GHA we use the Pushbot to track the latest run version per OS in a file in the repo. For TaskCluster, we don't have access to the Pushbot (probably), so we need to actually confer with TestRail.

### Type of change

- [x] Other Changes (Reporting)

The `reportable` fixture gets any relevant test runs associated with the current Version Under Test, and if any of them are populated, the fixture confirms that all the suites (less the number of non-reporting suites, currently `meta` and `pocket`) in the repo are covered in the report.

### Comments / Concerns

No way to force a report if the report already exists (say we want to re-run a report), even if tests were missed in one or more individual suites.


whlpatricia (approved):
Looks good!

PR #312, opened 2024-11-04
Anca/Print to PDF

soncuteanca:
### Description

- Verify that a webpage can be printed to PDF

### Bugzilla bug ID

**Testrail: https://mozilla.testrail.io/index.php?/cases/view/965142**
**Link: https://bugzilla.mozilla.org/show_bug.cgi?id=1925549**

### Type of change

- [ x] New Test

- In progress

- N/A

### Comments / Concerns

- I wasn't able to access the Print Preview destination dropdown, so instead, I set print.always_print_silent to True. This allows printing to proceed directly without the preview dialog. When no other printer is configured, the system defaults to PDF, allowing me to complete the test. Note that with this approach, the file name isn't automatically populated in the Save Output dialog. The File name field is left empty, so I had to enter it manually.

### DIFF
--- a/tests/printing_ui/test_print_to_pdf.py
+++ b/tests/printing_ui/test_print_to_pdf.py
@@ -2,10 +2,11 @@ import os
 from time import sleep
 
 import pytest
-from pynput.keyboard import Controller, Key
+from pynput.keyboard import Controller
 from selenium.webdriver import Firefox
 
 from modules.browser_object_panel_ui import PanelUi
+from modules.page_object_generics import GenericPdf
 
 
 @pytest.fixture()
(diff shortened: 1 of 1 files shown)


soncuteanca:
The issues with CI failures seem to be related to an existing problem affecting silent printing. The test passes on my Windows machine because the print_printer preference is set by default to Microsoft Print to PDF. However, if I switch to Mozilla Save to PDF (as it would be default on macOS), the test fails locally. The test consistently fails when using silent printing (with Mozilla Save to PDF default), both locally and in CI, on Mac and Windows but it runs without issues when I go through the Print Preview dialog manually. There are  some [open bugs]( https://bugzilla.mozilla.org/buglist.cgi?quicksearch=1879474%2C+1662471) to reflect silent printing issue.
While silent printing wasn’t an ideal solution to begin with, and i though it would be just a temporary solution to finish the test, given the unresolved issue it’s not a viable option at all.

ben-c-at-moz:
Accomplished with different PR

PR #316, opened 2024-11-06
Ben/tc confirm report

ben-c-at-moz:
The "confirm reportable" logic doesn't work because we attempted to use fixtures and hooks, needed to move the check to a pure module. This change should accomplish that move.


whlpatricia (approved):
lgtm!

PR #326, opened 2024-11-12
Add test for toggling visibility of the Bookmarks Toolbar

Tracy-Walker:
### Description
Adds test for toggling visibility of the bookmarks toolbar.

#### Bugzilla bug ID

**Testrail:** https://mozilla.testrail.io/index.php?/cases/view/2084637
**bug 1930233:**https://bugzilla.mozilla.org/show_bug.cgi?id=1930233

#### Type of change
- [X] New Test
- [X] Adds keyboard shortcut to toggle the Bookmarks Toolbar

Completes

### Comments 
Thanks to Patricia for discovering the collapsed attribute of the toolbar element


tests/bookmarks_and_history/test_toggle_bookmarks_toolbar.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def test_toggle_bookmark_toolbar(driver: Firefox):
+    """
+    C2084637: Verify that the user can Hide or Show the Bookmarks Toolbar using Keyboard shortcuts
+    """
+    # instantiate object
+    nav = Navigation(driver).open()
+
+    def is_bookmarks_toolbar_collapsed() -> str:
```
ben-c-at-moz, line 21: We could do this with:
```python
nav.element_attribute_contains("bookmarks-toolbar", "collapsed", "true")
```

This would eliminate the need for the hard waits as it uses `wait.until()` under the hood.

ben-c-at-moz (dismissed):
It's fine as is, but there is a way to eliminate the hard waits.

ben-c-at-moz (approved):
LGTM

PR #329, opened 2024-11-14
Hani/Open all bookmarks from bookmarks toolbar

sv-hyacoub:
### Description

Verify that the user can Open all the bookmarks from the Bookmarks Toolbar.

### Bugzilla bug ID

**Testrail: https://mozilla.testrail.io/index.php?/cases/view/2084564**
**Link: https://bugzilla.mozilla.org/show_bug.cgi?id=1930231**

### Type of change

- [x] New Test

Completed.

N/A.

### Comments / Concerns

N/A.


tests/bookmarks_and_history/test_open_all_bookmarks_from_bookmarks_toolbar.py
```diff
@@ -0,0 +1,44 @@
...
+    panel = PanelUi(driver)
+    context_menu = ContextMenu(driver)
+    page = GenericPage(driver)
+
+    # Have a few Bookmarks saved on the Toolbar
+    urls_to_bookmark = [URL1_TO_BOOKMARK, URL2_TO_BOOKMARK, URL3_TO_BOOKMARK]
+    for url in urls_to_bookmark:
+        GenericPage(driver, url=url).open()
+        nav.add_bookmark_via_star()
+
```
ben-c-at-moz, line 35: After this loop, we should load a page that we didn't bookmark, so we can ensure that we're not just picking up on that instance of the page.

ben-c-at-moz (changes requested):
One small thing, otherwise looks good!

tests/bookmarks_and_history/test_open_all_bookmarks_from_bookmarks_toolbar.py
sv-hyacoub: I added 2 lines to load a page that we didn't bookmark and to open bookmarks toolbar. Thanks

ben-c-at-moz (approved):

PR #331, opened 2024-11-15
code cleanup of .open() misusage and archive incident tests

Tracy-Walker:
### Description
Ben pointed out that usage of .open() in some cases is biting us.  This removes those unnecessary uses. 

#### Bugzilla bug ID
none

#### Type of change
- [X] Code cleanup

Have to see how this goes in CI

#### Comments / Concerns
I moved the incident tests into archived_tests directory.. Though pytest still finds them there.  I may just delete these test altogether,.


ben-c-at-moz (approved):
Big, big PR, but super necessary

PR #339, opened 2024-11-20
Hani/Stabilize test sap google adclick

sv-hyacoub:
### Description

Stabilize test sap google adclick.

### Bugzilla bug ID

**Testrail:**
**Link:**

### Type of change

Fixed.

N/A.

### Comments / Concerns

N/A.


tests/address_bar_and_search/test_sap_google_adclick.py
```diff
@@ -13,7 +13,6 @@ def test_case():
     return "1365108"
 
 
-@pytest.mark.unstable
```
ben-c-at-moz, line 16: Since this is failing in Linux, if you can't stabilize the test there, we can change the decorator to:
```python
import platform
@pytest.mark.xfail(platform.system() == "Linux", reason="Telemetry testing unstable in Linux")
```

ben-c-at-moz (changes requested):
If you get your Linux machine up in time, this would be worth debugging there. If not, go ahead with the suggested xfail marker

tests/address_bar_and_search/test_sap_google_adclick.py
sv-hyacoub: The test is failing on all OS's randomly if I run the test many times, sometimes search results don't return an ad and thats why the test can't see "search-result" locator.

ben-c-at-moz: In CI, it really looks like it's only failing in Linux--you're seeing fails in MacOS and Win?

Is there a way we can force there to be an ad, or re-run the check for SAP info multiple times without failing? Like run it 5 times and if it never gets anything it fails, but if it gets it once, it passes?

sv-hyacoub: Locally today the test passed everytime after running it up to 10 times on all 3 OS's, so I went with the first solution and mark it unstable on Linux.

ben-c-at-moz (approved):

PR #341, opened 2024-11-20
Remove test file with unstable tests

Tracy-Walker:
### Description
There is an **improved** version of this test file.  This simply removes the old unstable test file

#### Bugzilla bug ID
none

#### Type of change
- [X] Code cleanup

### Comments / Concerns
Once I enabled the old tests there apparently was conflict in Github CI with names.  We'll see how this PR does with the old file removed altogether


ben-c-at-moz (approved):
LGTM

PR #352, opened 2024-11-27
Pat/stabilize tests

whlpatricia:
### Description

test_about_logins_search is known to be stable
Did not run into issues for test_reader_view_location_bar

### Bugzilla bug ID

**Testrail:**
**Link:**

### Type of change

- [ ] New Test
- [ ] New POM

### Comments / Concerns


whlpatricia:
> let's keep an eye on these. They may not be that stable.

Yes, and if it fails in CI again we can have material to debug it :D

PR #355, opened 2024-12-02
Add mark for audio

Tracy-Walker:
### Description
Test uses audio, it should have @pytest.mark.audio

#### Bugzilla bug ID
https://bugzilla.mozilla.org/show_bug.cgi?id=1934627

### Type of change
- [X] Minor change

### Comments / Concerns
Test had been reported as "Untested" in TestRail.  Not sure if this is due to this missing mark, we'll see.


whlpatricia (approved):
lgtm

PR #356, opened 2024-12-02
Pat/added beta ver to env

whlpatricia:
### Description

Changed the way beta version number is tracked/committed so that there will not be issues when the version number is updated while there is an ongoing run.

### Bugzilla bug ID

**Testrail:**
**Link:** https://bugzilla.mozilla.org/show_bug.cgi?id=1934662

### Type of change

- [ ] New Test
- [ ] New POM
- [x] Other Changes (scheduled beta testing)

Fixed

NA

### Comments / Concerns

NA

### DIFF
--- a/tests/form_autofill/test_edit_credit_card.py
+++ b/tests/form_autofill/test_edit_credit_card.py
@@ -26,7 +26,8 @@ def hard_quit():
 tabs = [i for i in range(4)]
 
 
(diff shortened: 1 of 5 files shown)


tests/form_autofill/test_edit_credit_card.py
ben-c-at-moz, line 1: Is this test hanging now?

ben-c-at-moz (approved):

PR #362, opened 2024-12-05
Add code to wait for the test site to wake up

Tracy-Walker:
### Description
This adds a check for the page title to appear in the Tab before attempting to interact with the page. This has to be done as it often occurs the the site is "asleep" and must be allowed time to wake up.

#### Bugzilla bug ID
https://bugzilla.mozilla.org/show_bug.cgi?id=1935227

### Type of change
- [X] Test stabilization

Resolves the test stability issue


soncuteanca (approved):
lgtm

PR #368, opened 2024-12-05
fix linux test instability

ben-c-at-moz:
### Description

Stabilize previously failing test, TC has changed.


whlpatricia (dismissed):
lgtm

PR #376, opened 2024-12-11
remove unstable for tests where instability cannot be addressed atm

ben-c-at-moz:
### Description

QoL improvement where `unstable` is reserved for tests where:

* Test should not be run in CI (hang) / test likely to have negative effects
* Test should eventually work in our flow
* Test instability is not due to a failing dependency

These criteria eliminate the unstable mark in 6 test files.


whlpatricia (approved):
nice!

PR #393, opened 2024-12-23
Anca/Adjust blocking social media trackers test

soncuteanca:
### Description
- This test is blocked by bug [1866005](https://bugzilla.mozilla.org/show_bug.cgi?id=1866005), which is still open. However, the test xpassed each time, despite the Firefox feature being broken. This resulted in a false positive outcome. The settings were not correctly applied, so I resolved this by making sure  the settings are enabled and also, I've modified a condition in the tracking panel.  While the test appears intermittent, the consistent expected result should be xfail until a fix is landed.

### Bugzilla bug ID

- **Link: https://bugzilla.mozilla.org/show_bug.cgi?id=1938374**

### Type of change

- [x ] Other Changes (ensure the correct settings are enabled)

- Completed

- Completed

### Comments / Concerns

- I've tried numerous alternatives to replace the hard wait, including different condition waits and attribute checks, but none worked consistently. I'm uncertain about the root cause of this behavior, so for now, I've used a hard wait as a workaround.

### DIFF
--- a/modules/browser_object_navigation.py
+++ b/modules/browser_object_navigation.py
@@ -152,7 +152,7 @@ class Navigation(BasePage):
         """Open search settings from the awesome bar"""
         self.click_on("search-settings")
         return self
-    
+
     def click_on_change_search_settings_button(self) -> BasePage:
         with self.driver.context(self.driver.CONTEXT_CHROME):
             self.search_bar = self.find_element(By.CLASS_NAME, "searchbar-textbox")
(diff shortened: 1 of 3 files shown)


soncuteanca:
The test will fail due to [1866005](https://bugzilla.mozilla.org/show_bug.cgi?id=1866005), and as long as this bug remains open, it is likely to continue failing. This issue is reproducible on manual testing, as confirmed with two other colleagues today. The automated test required adjustments because the settings were not being properly recorded without a wait, which is why it consistently xpassed until now.

PR #396, opened 2025-01-02
Ben/reduce ci run

ben-c-at-moz:
* Create script that runs in CI flow to choose which tests to run in CI
  * If suite conftest is changed, run the whole suite. If main conftest or page base changed, run all tests. If object model changed, run all tests that import that model. If selector json changed, run all tests that import related model. If test changed, run that test.
* Modify CI flows to use that script on pull request
* Create case in the script for calls in scheduled Beta testing
* Update a number of tests to have `@pytest.mark.ci`, which means they run in the absence of any obvious changes.
* Lint lint lint


whlpatricia (approved):
great pr, this is amazing

whlpatricia (approved):
great!

whlpatricia (approved):
looks good to me

PR #402, opened 2025-01-07
always run ci headed tests

ben-c-at-moz:
Reduced CI has a bug where if no headed tests are changed, CI always fails. This fixes that by always running headed tests with the ci mark.


whlpatricia (approved):
lgtm!

PR #403, opened 2025-01-08
use getenv for ci selector script

ben-c-at-moz:
It appears we're not correctly identifying when a windows run is a scheduled beta, this one line _could_ fix that but there's only one way to find out.


whlpatricia (approved):
LGTM!

PR #405, opened 2025-01-08
Ben/fix win ci 2

ben-c-at-moz:
Previous attempt did not resolve the CI issue, this one should actually do it. We manage env vars completely differently in the windows flow (using pytest's built-in .env reader), so that env var wasn't set at the time the test picker script was executed.


ben-c-at-moz:
Force merging to try to get a sense for how this works

PR #413, opened 2025-01-17
disable google CAPTCHAd tests

Tracy-Walker:
### Description
Google has, as recently as FxBeta135 been identifying several of our test as bots and thus blocking those test with CAPTCHA.  This PR marks those tests as unstable for the time being.. We will most likely remove them from the STArFox repo and return this testing to our Manual team

#### Bugzilla bug ID
https://bugzilla.mozilla.org/show_bug.cgi?id=1938371

#### Type of change
- [X] Other Changes (disable tests with unstable mark)

paritially/temporarily resolves the issue

#### Comments / Concerns
We should remove these tests if we can't come up with another option


whlpatricia (approved):
Sad day in history

whlpatricia:
LGTM!

PR #414, opened 2025-01-17
always run all ci marked tests, to avoid unnecessary ci failures

ben-c-at-moz:
The CI subsuite isn't that long, we should just always run it, no matter what else is on the docket--this will prevent CI failures when there are changes to tests that don't get run in CI.


whlpatricia (approved):
lgtm!

PR #426, opened 2025-01-27
remove xfail mark in l10n tests

Tracy-Walker:
### Description
Since Ben fixed the Linux instability in CI form auto-fill tests, these marks can be removed from the test that got created in l10n_CM land.

#### Bugzilla bug ID
none

#### Type of change
- [X] Other Changes (remove xfail mark)


whlpatricia (approved):
you are right!

PR #434, opened 2025-02-03
Philimon[l10n-CM] Modularize l10n test suite

philimon-reset:
### Description

In order to avoid copy/pasting code and make the system as modular as possible, it is recommended that we create three folders under the L10n_CM folder in STARfox:

`unified`: which holds all the tests
`region`: which holds config files for the locales being tested
`constants`: which holds files that indicate selectors, text, and other test data that will not change

### Bugzilla bug ID: [1944613](https://bugzilla.mozilla.org/show_bug.cgi?id=1944613)

**Testrail:**
**Link:** [1Pager](https://docs.google.com/document/d/1-s-cDpXCZFYBJ9c0iwRgca3gM4nPhTL9gxTVd5Y_oto/edit?tab=t.0)

### Type of change

* This initially allows us to section the test suits more modularly. More work can be done though.
* Still need to find a way to allow reporting to and from test rails from the python script.
### Comments / Concerns
* Edited the tests in unified to allow them to run through `run_l10n.py`.
* User runs `python run_l10n.py US` to run all the US locale tests.
* Will return a `ValueError` if local is not listed in the `valid_locales` set. (could be moved from run_l10n to one of the constant file).

### DIFF
--- a/l10n_CM/Unified/conftest.py
+++ b/l10n_CM/Unified/conftest.py
@@ -1,12 +1,26 @@
+import os
+
 import pytest
 
 
 @pytest.fixture()
-def set_prefs():
-    """Set prefs"""
+def region():
+    return os.environ.get("STARFOX_REGION", "US")
+
+
+@pytest.fixture()
+def add_prefs(region: str):
     return [
         ("extensions.formautofill.creditCards.reauth.optout", False),
         ("extensions.formautofill.reauth.enabled", False),
         ("browser.aboutConfig.showWarning", False),
-        ("browser.search.region", "US"),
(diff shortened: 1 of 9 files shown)


ben-c-at-moz (approved):
Good work!

philimon-reset:
> this looks good. One request(doesn't block this landing); can we have a unified list of tests that applies to all regions and append any Region specific tests to the unified list? I ask, 'cause maintaining copies of lists for each region is going to get quite unwieldy.

The restructure can definitely be improved more, so organizing the tests as you suggested would be the next step.

PR #435, opened 2025-02-03
clean up excess code

ben-c-at-moz:
A few things have crept into the l10n suite that need to be cleaned up a little.


vsangereanMOZ (approved):
works ok.

PR #436, opened 2025-02-03
remove CA parametrization on autofill

ben-c-at-moz:
With the new l10n work, it doesn't make sense to double our coverage for autofill tests anymore


vsangereanMOZ (approved):
lgtm

PR #439, opened 2025-02-04
use collect_executables for version checking

ben-c-at-moz:
It was silly to check the version in two different scripts, so now we just check it in one script and use a flag to return the version number. We ALSO check one major version, one minor version, and one beta version up from what we believe is the latest. We ALSO make it so that we don't get tricked if a build is incomplete, and there's another one following it.


whlpatricia (approved):
LGTM!

PR #440, opened 2025-02-04
Include beta builds and improvements on reportable logic

whlpatricia:
### Description

Change existing logic so that the beta will be tested and reported if a new build within the beta is released. 
Improve logic such that github workflows for win and mac no longer require inputs and simply rely on collect_executables.py like linux in tc.
Removed tracking files as reportable works without implicitly tracking versions.

### Bugzilla bug ID

**Testrail:**
**Link:**

### Type of change

- [ ] New Test
- [ ] New POM

### Comments / Concerns


ben-c-at-moz (approved):
This looks good!

PR #442, opened 2025-02-05
Fix Scotch Bonnet broken test: test_add_engine_address_bar

Tracy-Walker:
### Description
Fixes Scotch Bonnet broken test; test_add_engine_address_bar

#### Bugzilla bug ID
https://bugzilla.mozilla.org/show_bug.cgi?id=1945160

#### Type of change
- [X] Test Stabilization

Completes


tests/downloads/test_download_pdf_from_context_menu.py
```diff
@@ -18,6 +18,7 @@ def delete_files_regex_string():
     return r".*i-9.pdf"
 
 
+@pytest.mark.unstable(reason="bug 1946131")
```
ben-c-at-moz, line 21: This should no longer fail with the changes from #444

tests/drag_and_drop/test_paste_image_text.py
```diff
@@ -53,6 +53,7 @@ def temp_selectors():
 COPY_URL = "https://1stwebdesigner.com/image-file-types/"
 
 
+@pytest.mark.unstable(reason="Bug 1938373")
```
ben-c-at-moz, line 56: This should no longer fail with the changes from #444

ben-c-at-moz (changes requested):

tests/drag_and_drop/test_paste_image_text.py
ben-c-at-moz: Yes, but if we mark it unstable, we'll stop running it in CI\

ben-c-at-moz: Actually, we should use `@pytest.mark.unstable` for flaky tests and stick to `.skip` for tests that never pass. You're right, I'll resolve these and approve the PR

ben-c-at-moz (dismissed):
LGTM, I'll make a separate PR to stop skipping "unstable" tests in CI/Scheduled Beta

ben-c-at-moz (dismissed):

ben-c-at-moz (approved):

ben-c-at-moz (approved):

ben-c-at-moz (dismissed):

whlpatricia (approved):
LGTM!

ben-c-at-moz:
gotta close then reopen the PR

PR #443, opened 2025-02-05
Ben/fix dupe tests ci

ben-c-at-moz:
The chooser script was put up against another unforeseen corner case and needed to be fixed. Tests that were made unstable by that change needed to be demarked.


ben-c-at-moz:
Covered by #444

PR #444, opened 2025-02-06
Ben/fix audio video scotch bonnet

ben-c-at-moz:
Fix the AV test that was failing.

The chooser script was put up against another unforeseen corner case and needed to be fixed. Tests that were made unstable by that change needed to be demarked.


whlpatricia (approved):
LGTM!

PR #445, opened 2025-02-06
Philimon/suite refactor

philimon-reset:
### Description
General Improvement of the `find_toolbar` and `pdf_viewer` test suites.
This includes improvements to the `GenericPDF`, `Navigation`, and `FindToolbar` object models.

### Bugzilla bug ID: 1935748

**Testrail:**
**Link:** [Points of Improvment.pdf](https://github.com/user-attachments/files/18691355/Points.of.Improvment.pdf)

### Type of change

Clean up to improve the tests' and object models' readability and clarity.

### Comments / Concerns
Test improvements include clean-up, moving context switches to decorators, moving large functionalities to object models, and more.

### DIFF
--- a/modules/browser_object_find_toolbar.py
+++ b/modules/browser_object_find_toolbar.py
@@ -20,6 +20,7 @@ class FindToolbar(BasePage):
             self.wait_for_page_to_load()
         return self
 
+    @BasePage.chrome_context(self.driver)
     def open_with_key_combo(self) -> BasePage:
         """Use Cmd/Ctrl + F to open the Find Toolbar, wait for load"""
         with self.driver.context(self.driver.CONTEXT_CHROME):
--- a/modules/browser_object_find_toolbar.py
+++ b/modules/browser_object_find_toolbar.py
@@ -1,6 +1,7 @@
 import json
 import logging
 
+from selenium.webdriver import Firefox
 from selenium.webdriver.common.keys import Keys
 from selenium.webdriver.support import expected_conditions as EC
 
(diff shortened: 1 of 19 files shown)


modules/browser_object_find_toolbar.py
```diff
@@ -11,79 +12,95 @@
...
-            findbar = self.get_element("find-toolbar-input")
-            findbar.click()
-            findbar.clear()
-            findbar.send_keys(term + Keys.ENTER)
+        find_bar = self.get_element("find-toolbar-input")
+        find_bar.click()
+        find_bar.clear()
+        find_bar.send_keys(term + Keys.ENTER)
+        if find_bar.get_property("status") != "notfound":
+            self.match_dict = self.get_match_args()
```
ben-c-at-moz, line 47: What are we doing with `match_dict`, in the grand scheme of things?

tests/find_toolbar/conftest.py
```diff
@@ -10,3 +14,15 @@ def suite_id():
...
     return []
+
+
+@pytest.fixture()
+def find_toolbar(driver: Firefox):
+    find_toolbar = FindToolbar(driver)
+    yield find_toolbar
+
+
+@pytest.fixture()
```
ben-c-at-moz, line 25: This may be for much later but...we might want to do away with `BrowserActions` at some point, and either way, I probably shouldn't have gotten in the habit of calling it `ba` in tests. So maybe we'll work on this part a bit.

modules/page_base.py
```diff
@@ -103,6 +104,28 @@ def set_chrome_context(self):
         if self._xul_source_snippet not in self.driver.page_source:
             self.driver.set_context(self.driver.CONTEXT_CHROME)
 
+    @staticmethod
```
ben-c-at-moz, line 107: Awesome way of solving the problem.

tests/find_toolbar/test_find_in_pdf.py
```diff
@@ -12,7 +12,7 @@ def test_case():
 
 
 WORD_SEGMENT = "authori"
-MATCH_TWO_COTEXT = "representative must complete"
+MATCH_TWO_CONTEXT = "representative must complete"
```
ben-c-at-moz, line 15: Ah, I think I was too much of a dork here, because "co-text" is actually what I meant. We could call it `MATCH_TWO_COMPLETION` or something

tests/find_toolbar/test_find_toolbar_nav.py
```diff
@@ -16,58 +15,57 @@ def test_case():
...
+PROCESSES_SELECTOR = "a[href='about:processes']"
 
-def compare(a: int, b: int) -> bool:
-    return abs(a - b) < TOLERANCE
+first_match = (By.CSS_SELECTOR, DELETE_PROFILE_SELECTOR)
+fourth_match = (By.CSS_SELECTOR, PROCESSES_SELECTOR)
 
 
-def test_find_toolbar_navigation(driver: Firefox):
+def are_lists_different(a: int, b: int) -> bool:
```
ben-c-at-moz, line 26: this is a good rename

tests/find_toolbar/test_find_toolbar_nav.py
```diff
@@ -16,58 +15,57 @@ def test_case():
...
+    return abs(a - b) > TOLERANCE
```
ben-c-at-moz, line 27: do we want `>` or `>=`? I'm legit not sure.

tests/find_toolbar/test_find_toolbar_nav.py
```diff
@@ -16,58 +15,57 @@ def test_case():
...
+
+
+def test_find_toolbar_navigation(
+    driver: Firefox, find_toolbar: FindToolbar, ba: BrowserActions
+):
     """
     C127249: Navigate through found items
+
+    Arguments:
+        ba: instantiation of BrowserActions BOM.
```
ben-c-at-moz, line 37: BrowserActions isn't strictly a BOM. Let's talk about BrowserActions sometime, I do think it might be worth considering eliminating it and merging it with BasePage

tests/pdf_viewer/test_download_pdf_with_form_fields.py
```diff
@@ -2,10 +2,13 @@
 from time import sleep
 
 import pytest
+from pynput.keyboard import Controller
```
ben-c-at-moz, line 5: So I know that it's standard to import everything at the top, but in the case of pynput, we really don't want to do that--the simple act of importing pynput on a *nix system without a DISPLAY env var set causes a crash.

tests/pdf_viewer/test_pdf_download.py
```diff
@@ -22,32 +25,45 @@ def delete_files_regex_string():
...
     fillable_pdf_url: str,
     downloads_folder: str,
     sys_platform,
     delete_files,
+    file_name,
 ):
     """
     C3932: PDF files can be successfully downloaded via pdf.js
+
+
```
ben-c-at-moz, line 45: Let's also talk about whether we need args in the docstring for tests. I can see it both ways, and I want to decide what are the criteria for inclusion in the docstring.

tests/find_toolbar/conftest.py
philimon-reset: That makes a lot of sense. Most methods can be moved to the `page_base` or the corresponding `BOM.` For now, I can rename the instance fixture to be more descriptive of what it is.

tests/find_toolbar/test_find_toolbar_nav.py
philimon-reset: I think it should be `>` since anything beyond the `TOLERANCE` would be a different list.

tests/find_toolbar/test_find_toolbar_nav.py
philimon-reset: Definitely Agree

tests/pdf_viewer/test_pdf_download.py
philimon-reset: Thats fair. I think some comments on the args are necessary, especially until we reach a point where we all have descriptive parameters. But we could skip the repeated parameters like `driver` since we all know what it is.

tests/pdf_viewer/test_download_pdf_with_form_fields.py
philimon-reset: I suspected there was a reason the import wasn't at the top. my bad, should have asked before the change 😅

tests/find_toolbar/test_find_in_pdf.py
philimon-reset: Just changed it 👍🏽

tests/pdf_viewer/test_zoom_pdf_viewer.py
```diff
@@ -16,45 +16,61 @@ def test_case():
...
+
+    Arguments:
+        control: zoom controls
+        fillable_pdf_url: pdf file directory path
     """
-    pdf_page = GenericPdf(driver, pdf_url=fillable_pdf_url).open()
+    zoom_pdf_viewer(driver, fillable_pdf_url, control, "Keys")
 
+
+def zoom_pdf_viewer(
```
ben-c-at-moz, line 39: Maybe this should move to the GenericPdf POM or to the pdf_viewer conftest? Seems like a generic function that may be needed in future tests...

tests/pdf_viewer/test_pdf_input_numbers.py
```diff
@@ -29,17 +29,21 @@ def test_pdf_input_numbers(
...
-    numeric_field = pdf.get_element("zipcode-field")
-
-    # Test value to input in the field
-    test_value = "12345"
 
     # Clear the field and enter the test value
-    numeric_field.send_keys(test_value + Keys.TAB)
+    numeric_field = pdf.get_element("zipcode-field")
+    numeric_field.clear()
+    numeric_field.send_keys("12345" + Keys.TAB)
```
ben-c-at-moz, line 46: is there a reason to move the constant test value out of the variable and into two different lines? Open to it, but would love to know the rationale

tests/pdf_viewer/test_pdf_download.py
ben-c-at-moz: Yeah, that's a good point

ben-c-at-moz:
Something went wrong in the CI checks, and the tests were hanging. Often this happens because a context menu was left open (when automation is running via selenium, context menus do not close on their own).

ben-c-at-moz (dismissed):
Good job, address those merge conflicts and we're good to go!

ben-c-at-moz (approved):

PR #468, opened 2025-02-13
Hani/Demo address data captured in doorhanger and stored

sv-hyacoub:
### Description

Verify Address data are captured in the Capture Doorhanger and stored in about:preferences

### Bugzilla bug ID

**Testrail: https://mozilla.testrail.io/index.php?/cases/view/2888703**
**Link: https://bugzilla.mozilla.org/show_bug.cgi?id=1943171**

### Type of change

- [x] New Test

Completed.

N/A

### Comments / Concerns

N/A

### DIFF
--- a/SELECTOR_INFO.md
+++ b/SELECTOR_INFO.md
@@ -1353,6 +1353,48 @@ Description: Save address doorhanger
 Location: Address bar
 Path to .json: modules/data/autofill_popup.components.json
 ```
+```
+Selector Name: address-doorhanger-street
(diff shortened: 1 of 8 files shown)


soncuteanca (approved):
I initially thought the long dictionary could be avoided by defaulting faker state to abbreviations. However, since a dictionary is still necessary (although shorter), I think your approach is better, by keeping full names as the general rule and handling abbreviations separately in tests. LGTM

PR #470, opened 2025-02-14
Remove skipif to see if these are working in CI

Tracy-Walker:
### Description
A couple of tests got marked with skip if on Linux due to "136.0b3 Linux security bustage". Re-enabling them to see if they'll pass CI now

### Bugzilla bug ID
NA

### Type of change
- [X] skipif mark removal


ben-c-at-moz (approved):
LGTM

PR #472, opened 2025-02-18
Philimon/Create test for Verify that a new Address can be added

philimon-reset:
### Description

- Adds a new test to verify addition in saved addresses for US region.
- Minor changes in the other tests in the l10n test suite

### Bugzilla bug ID

**Testrail:** [2886580](https://mozilla.testrail.io/index.php?/cases/view/2886580&group_by=cases:section_id&group_order=asc&display_deleted_cases=0&group_id=579542)
**Link:**

### Type of change

- [x] New Test

### Comments / Concerns

- Added a test that create a saved address entry and verifies the address for US region.
- move class instantiation in the other tests in l10n to the `contest`.
- removed any explicit region change in the test itself since the region was specified in the `conftest`. 
- added some additional methods in `AboutPrefs` to streamline the process of using the saved addresses.

### DIFF
--- a/l10n_CM/Unified/conftest.py
+++ b/l10n_CM/Unified/conftest.py
@@ -2,6 +2,11 @@ import os
 
 import pytest
 
+from modules.browser_object_autofill_popup import AutofillPopup
+from modules.page_object_autofill import AddressFill, CreditCardFill
+from modules.page_object_prefs import AboutPrefs
+from modules.util import Utilities
+
 
 @pytest.fixture()
 def region():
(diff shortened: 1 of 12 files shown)


l10n_CM/Unified/test_demo_ad_address_data_captured_in_doorhanger_and_stored.py
```diff
@@ -13,17 +13,16 @@ def test_case():
 
 
 def test_demo_ad_address_data_captured_in_doorhanger_and_stored(
```
ben-c-at-moz, line 15: is this `_ad_` or `_add_`?

l10n_CM/region/US.json
```diff
@@ -6,8 +6,7 @@
         "test_demo_ad_doorhanger_shown_on_valid_address_submission.py",
```
ben-c-at-moz, line 6: Seems to me that all of these tests are in common between the regions. I think we should have another `.json` file, something like `common.json` or `unified.json` where all these test names live, then update `US.json` and the other regional JSONs to have empty lists for `tests` at the moment, until we have tests that are for individual regions only. You'll need to update the runner script too.

l10n_CM/Unified/conftest.py
```diff
@@ -19,8 +24,38 @@ def add_prefs(region: str):
 
 
 @pytest.fixture()
-def set_prefs(add_prefs: dict):
+def set_prefs(add_prefs: dict, region: str):
```
ben-c-at-moz, line 27: Seems like, again we can factor out the common elements. I would move the prefs from `add_prefs` down here in the `prefs = [` line, again having `add_prefs` return an empty list for now. Then if individual tests need additional prefs, we can create `add_prefs` fixtures inside the tests. If we find that these prefs are not actually common among all future l10n tests we can cross that bridge (perhaps implementing `overwrite_prefs` or something else.

ben-c-at-moz (changes requested):
A couple of structural things, but the code is good.

ben-c-at-moz (dismissed):
LGTM

PR #473, opened 2025-02-18
Attempt to fix the "ghost runs" issue in scheduled beta

ben-c-at-moz:
Also, update Ruff to newest version and re-lint. Sorry!


whlpatricia (approved):
LGTM

PR #476, opened 2025-02-19
retry collect executables url on failure

ben-c-at-moz:
Collect executables bombs out weirdly if the FTP site is not responsive. Retry and give an error.


philimon-reset (approved):
LGTM

PR #480, opened 2025-02-21
Hani/ Test demo ad autofill address fields

sv-hyacoub:
### Description

Verify Autofill functionality when selecting an entry from the dropdown for address fields.

### Bugzilla bug ID

**Testrail: https://mozilla.testrail.io/index.php?/cases/view/2888563**
**Link: https://bugzilla.mozilla.org/show_bug.cgi?id=1943174**

### Type of change

- [x] New Test

Need to find a solution to improve normalize_phone_number since it’s covering only US and CA. Sometimes DE and FR tests are failing due to not match phone number.
The function always trims to the last 10 digits, even if the country code is included in the number. For FR/DE, this can lead to incorrect truncation.

N/A

### Comments / Concerns

N/A


l10n_CM/Unified/test_demo_ad_autofill_address_fields.py
```diff
@@ -0,0 +1,63 @@
...
+from modules.page_object_autofill import AddressFill
+from modules.util import Utilities
+
+
+@pytest.fixture()
+def test_case():
+    return "2888563"
+
+
+def autofill_and_verify(address_autofill, address_autofill_popup, field_label, address_autofill_data, region, util):
```
soncuteanca, line 14: Maybe this method can be moved in POM, and used as well in the other two tests (although those tests covers only 2 fields (name/organisation, respectively email/tel).

soncuteanca (dismissed):

l10n_CM/Unified/test_demo_ad_autofill_address_fields.py
```diff
@@ -0,0 +1,63 @@
...
+    address_autofill_popup.click_doorhanger_button("save")
+
+    # List of field labels to be autofilled and verified
+    fields_to_test = [
+        "street-address",
+        "address-level2",
+        "address-level1",  # This will be skipped for DE/FR
+        "postal-code",
+        "country",
+        "email",
```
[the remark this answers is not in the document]
sv-hyacoub: thanks, I just removed email and tel from here.

l10n_CM/Unified/test_demo_ad_autofill_address_fields.py
sv-hyacoub: I moved this to POM. Thanks.

l10n_CM/Unified/test_demo_ad_autofill_address_fields.py
```diff
@@ -0,0 +1,44 @@
...
+        "street-address",
+        "address-level2",
+        "address-level1",  # This will be skipped for DE/FR
+        "postal-code",
+        "country"
+    ]
+
+    # Loop through each field and perform the autofill test
+    for field in fields_to_test:
+        address_autofill.autofill_and_verify(address_autofill, address_autofill_popup, field,
```
philimon-reset, line 42: you shouldn't need pass `address_autofill` since `autofill_and_verify` is a method of `AddressFill`. you can access `verify_autofill_data` and `double_click` using `self`.

philimon-reset, line 42: I also don't think `region` is needed since you are already passing in `address_autofill_data` you could be able to access the region through the `country` attribute from `address_autofill_data`

l10n_CM/Unified/test_demo_ad_autofill_address_fields.py
```diff
@@ -0,0 +1,44 @@
...
+def test_case():
+    return "2888563"
+
+
+def test_demo_ad_autofill_address_fields(driver: Firefox, region: str):
+    """
+    C2888563 - Verify Autofill functionality when selecting an entry from the dropdown for address fields
+    """
+    # Instantiate objects
+    address_autofill = AddressFill(driver)
```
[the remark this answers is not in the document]
sv-hyacoub: Done, thanks.

l10n_CM/Unified/test_demo_ad_autofill_address_fields.py
sv-hyacoub: Done, thanks.

sv-hyacoub: I'm not sure how to do that since I'm still using region in autofill_and_verify, I don't know how to access country attribute in this case.

ben-c-at-moz (changes requested):
Code looks good, and works, but the test needs to be added to Unified.json

ben-c-at-moz (approved):
Good to go!

PR #485, opened 2025-02-25
Philimon/L10N_Script_Modification

philimon-reset:
### Description

Modification to `run_l10n.py` to allow for flags and running of unified tests.

### Bugzilla bug ID

**Testrail:**
**Link:**

### Type of change

### Comments / Concerns
This update ensures that Unified tests always run for every requested region without requiring the `run_l10n.py Unified` command. Now, when executing `run_l10n.py US`, it will first run all Unified tests in the US region, followed by all US-specific tests. If no region is specified, the script will automatically run all Unified tests across all supported regions, followed by any region-specific tests where applicable (currently none). Additionally, support for the flags` --run-headless`, `-n`, and `--reruns` has been incorporated.


ben-c-at-moz (commented):
So, the code looks fine, but I'm noting that some of the tests don't seem to be running in the region specified, specifically these ones:

```
l10n_CM/Unified/test_demo_ad_verify_new_address_added.py::test_verify_new_address_is_added
l10n_CM/Unified/test_demo_cc_add_new_credit_card.py::test_create_new_cc_profile
l10n_CM/Unified/test_demo_cc_clear_form.py::test_cc_clear_form
l10n_CM/Unified/test_demo_cc_doorhanger_data_is_stored_in_about_prefs.py::test_demo_cc_data_captured_in_doorhanger_and_stored
l10n_CM/Unified/test_demo_cc_doorhanger_shown_on_valid_credit_card_submission.py::test_cc_check_door_hanger_is_displayed
l10n_CM/Unified/test_demo_cc_dropdown-presence.py::test_dropdown_presence_credit_card
l10n_CM/Unified/test_demo_cc_yellow_highlight.py::test_cc_yellow_highlight
l10n_CM/Unified/test_demo_ad_yellow_highlight.py::test_address_yellow_highlight_on_name_organization_fields
```

If these tests aren't supposed to run with localized data, they should probably not be in the unified category. If they are, I'm not sure that FR data is being used when they're being tested in FR

philimon-reset:
> test_demo_cc_yellow_highlight

You're correct. The tests either don't use localized data or are limited to only a specific region.
- I moved `test_demo_ad_verify_new_address_added.py` to the `US` region since it is specific. However, we will move it to unified once I have finished the issue with the globalized address filling.
- All the `cc` tests are region-independent, so I moved them to a new category called `Isolated.` This would include any tests that don't use localized data. I Wasn't sure if we should run them regardless of the options given to the runner script, so I added an option called `-isolated` to run the tests.
   - I added the default region for the logging to be `US`, but I can change this if necessary. 
- `test_demo_ad_yellow_highlight.py` is not strictly dependent on a region for the test assertion but requires the region to be specified for the faker data, so I left it `unified`.

l10n_CM/Unified/conftest.py
```diff
@@ -11,7 +11,7 @@
 
 @pytest.fixture()
 def region():
-    return os.environ.get("STARFOX_REGION", "US")
+    return os.environ.get("FX_REGION", "FR")
```
ben-c-at-moz, line 14: testing

ben-c-at-moz (dismissed):
We talked through all this, and it looks great now. Thank you for sticking with it.

philimon-reset:
> We talked through all this, and it looks great now. Thank you for sticking with it.



> We talked through all this, and it looks great now. Thank you for sticking with it.

No worries 😄

PR #488, opened 2025-02-27
Unify opening browser windows.

Tracy-Walker:
### Description
Attempt at being able open a new browser window; tab, window or private window from one method. 

#### Bugzilla bug ID

[Bug 1049884](https://bugzilla.mozilla.org/show_bug.cgi?id=1949884)

#### Type of change
- [X] Stability and convenience

I think I found all uses.  Many instances of opening a new windows are specifically testing various context menus and the like.  As such, this was most picking up opening new private windows

#### Comments / Concerns
There may be tests I missed. More importantly there was some serious ghost in the machine type stuff happening developing/debugging this. Not confident it will pass CI or keep passing in CI.  :/

### DIFF
--- a/modules/page_base.py
+++ b/modules/page_base.py
@@ -703,12 +703,12 @@ class BasePage(Page):
             browser_window: Can be a standard 'window', 'tab' or 'private' browser window.
         """
         if browser_window == "private":
-            self.open_and_switch_to_private_window()
+            self.open_and_switch_to_private_window_via_keyboard()
         else:
             self.driver.switch_to.new_window(browser_window)
         return self
 
-    def open_and_switch_to_private_window(self) -> Page:
+    def open_and_switch_to_private_window_via_keyboard(self) -> Page:
         """
         Opens a new private browsing window via keyboard shortcut and switch to it
         """


modules/page_base.py
```diff
@@ -678,20 +679,55 @@ def wait_for_num_tabs(self, num_tabs: int) -> Page:
...
+        Parameters:
+            browser_window: Can be a standard 'window', 'tab' or 'private' browser window.
+        """
+        if browser_window == "private":
+            self.open_and_switch_to_private_window()
+        else:
+            self.driver.switch_to.new_window(browser_window)
+        return self
+
+    def open_and_switch_to_private_window(self) -> Page:
```
ben-c-at-moz, line 711: So it seems we have `BasePage.open_and_switch_to_new_window("private")` and `BasePage.open_and_switch_to_private_window()`, which both do the same thing, but by different methods. I suspect this will get confusing in the future. (I recognize we have done things like this in the past, but it's probably a habit we want to break.) I would consider renaming this method to make it explicit that we're opening a private window with the key shortcut. It's also possible that we want to move it out of the BasePage, but I'm not 100% set on that.

modules/page_base.py
```diff
@@ -678,20 +679,55 @@ def wait_for_num_tabs(self, num_tabs: int) -> Page:
...
-        self.driver.switch_to.window(self.driver.window_handles[-1])
+        with self.driver.context(self.driver.CONTEXT_CONTENT):
+            self.driver.switch_to.window(self.driver.window_handles[-1])
         return self
 
     def switch_to_new_window(self) -> Page:
         """Switch to the most recently opened window. Can be a standard or private window"""
-        all_window_handles = self.driver.window_handles
-        self.driver.switch_to.window(all_window_handles[-1])
+        with self.driver.context(self.driver.CONTEXT_CONTENT):
```
ben-c-at-moz, line 688: This feels pretty stable, is context the whole reason it was flaky before?

ben-c-at-moz (changes requested):
I have one structural concern, but it's mostly a naming thing. The rest of the code looks fine.

ben-c-at-moz (approved):
Looks good, I think this is a big improvement over the previous state of affairs.

PR #489, opened 2025-02-27
Hani/l10n ad - clear form from name and org

sv-hyacoub:
### Description

Verify clear functionality after selecting an entry from name/org fields.

### Bugzilla bug ID

**Testrail: https://mozilla.testrail.io/index.php?/cases/view/2888560**
**Link: https://bugzilla.mozilla.org/show_bug.cgi?id=1943170**

### Type of change

- [x] New Test

Completed.

N/A

### Comments / Concerns

N/A


ben-c-at-moz (changes requested):
Test needs to be added to Unified.json, but the code looks good and it works!

ben-c-at-moz (dismissed):

soncuteanca (approved):
Looks good! Not sure if a common method would work, as the approach differs slightly between credit cards and addresses. However, we can revisit this later and give it a try.

PR #494, opened 2025-02-28
tracy/ Standardize method to set and add prefs

Tracy-Walker:
### Description
Refactor of the entire STArFox project to have every conftest.py and involved tests use the same method of setting and adding preferences

#### Bugzilla bug ID

[Bug 1948358](https://bugzilla.mozilla.org/show_bug.cgi?id=1948358)

#### Type of change
- [X] Code conssitency

Completes

#### NOTES:
Going forward all test suite conftest.py should have the new template for prefs_list and add_to_prefs_list fixtures.  
Individual tests always should use add_to_prefs_list when needed to set a pref just for that test.

### DIFF
--- a/tests/password_manager/test_password_csv_export.py
+++ b/tests/password_manager/test_password_csv_export.py
@@ -19,6 +19,10 @@ MAC_GHA = environ.get("GITHUB_ACTIONS") == "true" and sys.platform.startswith("d
 
 @pytest.mark.headed
 @pytest.mark.skipif(MAC_GHA, reason="Test unstable in MacOS Github Actions")
+@pytest.mark.skipif(
+    sys.platform.lower().startswith("linux"),
+    reason="Test unstable in CI on Linux. see bug https://bugzilla.mozilla.org/show_bug.cgi?id=1951666",
+)
 def test_password_csv_export(
     driver_and_saved_logins, home_folder, sys_platform, opt_ci
 ):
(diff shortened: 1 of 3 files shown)


ben-c-at-moz:
I'll check this out hopefully today, but please do not merge until 137.0b1 is tested on 2025-03-03. Thanks!

tests/password_manager/test_password_csv_export.py
```diff
@@ -13,7 +15,11 @@ def test_case():
     return "2241521"
 
 
+MAC_GHA = environ.get("GITHUB_ACTIONS") == "true" and sys.platform.startswith("darwin")
```
ben-c-at-moz, line 18: Same question as above, just a little worried when tests go unstable...

tests/preferences/test_firefox_home_on_launch.py
```diff
@@ -12,7 +12,7 @@ def test_case():
 
 
 @pytest.fixture()
-def set_prefs():
+def add_to_prefs_list():
     """Set prefs"""
```
ben-c-at-moz, line 16: Lint: comment no longer applies

tests/address_bar_and_search/test_search_code_google_non_us.py
ben-c-at-moz, line 1: Might be able to stabilize this test even, I think the code should be `client=firefox-b-1-d` instead of `client=firefox-b-d`

ben-c-at-moz (changes requested):
The only changes I'm actually strongly requesting are lint, the rest are more questions than issues. Thanks for doing this, bet it was a pain.

tests/password_manager/test_password_csv_export.py
ben-c-at-moz: If you don't mind adding a comment to that effect in the code so we don't forget the reason, that would be great.

ben-c-at-moz (approved):
LGTM

PR #496, opened 2025-02-28
prep workflow for differently-named Fx executables in Mac

ben-c-at-moz:
When we test devedition, the MacOS executable is called "Firefox Developer Edition", which matters as `hdiutil` mounts the volume so we can tell Selenium where Fx lives. This change *should* allow that to happen while also preventing scheduled Beta from breaking.

* `parse_executables` parses the output of `hdiutil` mount to get the name of the package
* pipeline updated to use that script to construct `FX_EXECUTABLE`
* Change cron in devedition gha yaml
* Update testrail_integration.collect_changes() not to break on missing `FX_CHANNEL`

### DIFF
--- a/.github/workflows/check-devedition.yml
+++ b/.github/workflows/check-devedition.yml
@@ -2,7 +2,7 @@ name: Check new devedition version
 
 on:
   schedule:
-    - cron: "38 * * * *"
+    - cron: "38 */12 * * *"
 env:
   LATEST: ""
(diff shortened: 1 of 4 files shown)


ben-c-at-moz:
Please do not merge until 137.0b1 is tested on 2024-03-03

whlpatricia (approved):
LGTM!

PR #500, opened 2025-03-03
fix private window race condition, update PR template

ben-c-at-moz:
### Changes

* Update private window method in BasePage to `driver.get("about:blank")` before returning, I believe something in Selenium's switch_to isn't fully finishing up before returning, causing the window not to be able to navigate when we ask it to
* Update PR template, we're not just pumping out tests and POMs anymore

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,28 +1,37 @@
-### Description
+### Relevant Links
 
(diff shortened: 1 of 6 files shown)


ben-c-at-moz:
Something weird happened with git, broke this PR.

PR #508, opened 2025-03-06
Philimon/hover verify

philimon-reset:
### Relevant Links

Bugzilla: [1943173](https://bugzilla.mozilla.org/show_bug.cgi?id=1943173)
TestRail: [2888562](https://mozilla.testrail.io/index.php?/cases/view/2888562)

Added ability to test previewed data on hovered fields and verify them.

### Process Changes Required

_Mark the relevant boxes:_

- [x] Changes or creates a BOM/POM (name the object model): AutoFill

To verify the previewed data that appeared when the autofill field is hovered, I had to pull the data from the chrome context as that is where the data for the preview is stored. I had problems with making it work for all regions but since DE and FR don't save address-line1 I had to add a small hack until a better solution is found.
### Comments or Future Work

Will find a way to remove the condition check for DE and FR and make it region agnostic.

### Workflow Checklist

- [x] Please request reviewers
- [x] If this is an unblocker, please post in Slack.
- [ ] If asked to address comments, please resolve conversations.
- [ ] If asked to change code, please re-request review from the person who wanted changes.

Thank you!

### DIFF
--- a/modules/page_base.py
+++ b/modules/page_base.py
@@ -684,6 +684,11 @@ class BasePage(Page):
         self.driver.switch_to.default_content()
         return self
 
+    def switch_to_iframe(self, index: int):
+        """Switch to frame of given index"""
+        self.driver.switch_to.frame(index)
+        return self
+
     def switch_to_new_tab(self) -> Page:
         """Get list of all window handles, switch to the newly opened tab"""
         with self.driver.context(self.driver.CONTEXT_CONTENT):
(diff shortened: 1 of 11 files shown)


modules/page_base.py
```diff
@@ -708,6 +708,11 @@ def switch_to_default_frame(self) -> Page:
         self.driver.switch_to.default_content()
         return self
 
+    def switch_to_iframe(self, index: int):
```
ben-c-at-moz, line 711: We now have `switch_to_frame()` and `switch_to_iframe()` with different implicit expectations. Maybe rename this to `switch_to_frame_by_index()`?

ben-c-at-moz (approved):
One tiny thing, we'll go ahead and merge and I'll make a new PR for this small change, because we need to unblock

modules/page_base.py
ben-c-at-moz: I'll do this in a new PR

PR #523, opened 2025-03-17
vs/Adapt the tests to work with Firefox, Nightly and Devedition

vsangereanMOZ:
### Relevant Links

Bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1954091 and https://bugzilla.mozilla.org/show_bug.cgi?id=1954094

Changed the tests to check only for the word "Firefox"


vsangereanMOZ:
I don't understand why this happened. I made a checkout main and git pull before.

PR #530, opened 2025-03-19
Create TestRail change custom field script

Tracy-Walker:
#### Description of Code / Doc Changes

TestRail task

####  Explanations
See comments in the script

### Comments 
We don't merge this.  It's already been run against the 3 suites assigned to me.

### DIFF
--- a/modules/testrail_change_custom_fields.py
+++ b/modules/testrail_change_custom_fields.py
@@ -21,8 +21,8 @@ from testrail import TestRail
 
 # Configuration - Update these values accordingly
 TESTRAIL_HOST = "https://mozilla.testrail.io"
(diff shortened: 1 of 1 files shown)


ben-c-at-moz (commented):
This is good--and I checked TestRail for any inadvertent status changes, and it looks like there are none.

ben-c-at-moz:
> No that it matters for this PR (as it won't be landed), Linux failed in trying to fix ruff?

Yeah, if you import something at the module level, ruff wants it to be at the top of the file...and it's one of the few things that ruff considers unsafe to fix on its own. It's not a major concern for this project, of course, but if it were something we wanted to merge, we'd have to --fix-unsafe or move it ourselves.

PR #534, opened 2025-03-21
tracy/Disable AltText

Tracy-Walker:
#### Relevant Links
Bugzilla: [https://bugzilla.mozilla.org/show_bug.cgi?id=1953903](https://bugzilla.mozilla.org/show_bug.cgi?id=1953903)

#### Description of Code
Add additional pref change to disable the add alt text to pdf image feature

#### Workflow Checklist
- [X] Please request reviewers

### DIFF
--- a/tests/pdf_viewer/test_add_image_pdf.py
+++ b/tests/pdf_viewer/test_add_image_pdf.py
@@ -39,7 +39,7 @@ def test_add_image_pdf(driver: Firefox, sys_platform, pdf_viewer: GenericPdf):
 
     Arguments:
         sys_platform: Current System Platform Type
-        pdf_viewer: instance of GenericPdf with correct path.
+        pdf_viewer: Instance of GenericPdf with correct path.
     """
     context_menu = ContextMenu(driver)
     image_path = Path("data") / IMAGE_FILE_NAME


modules/page_object_generics.py
```diff
@@ -77,27 +77,29 @@ def add_image(self, image_path: str, sys_platform: str) -> BasePage:
...
             keyboard.type(image_path)
         sleep(1)
         keyboard.press(Key.enter)
         keyboard.release(Key.enter)
-        sleep(2)
+        sleep(1)
         keyboard.press(Key.enter)
         keyboard.release(Key.enter)
-        sleep(2)
+        sleep(1.5)
```
ben-c-at-moz, line 97: so this was previously added to cycle through the fields of the Alt Text add screen. I should have documented that, and that's my bad. So this may be what's causing intermittent hanging in the Linux builds

ben-c-at-moz: and by "this", I mean the three tabs below. We shouldn't need them anymore

ben-c-at-moz (changes requested):
I think there's some code that needs to be removed, but other than that, it's good.

ben-c-at-moz:
The hang in the previous check appears to be due to a worker being improperly initialized, i.e. a taskcluster glitch. Good to merge!

PR #551, opened 2025-04-08
vs/Part1-Refactoring

vsangereanMOZ:
### Relevant Links

Bugzilla: [1953253](https://bugzilla.mozilla.org/show_bug.cgi?id=1953253)
TestRail: _

Code improvemenets for tests under Adress Bar and Search.


tests/address_bar_and_search/test_addon_suggestion.py
```diff
@@ -35,14 +32,18 @@ def test_addon_suggestion_based_on_search_input(driver: Firefox):
 
     nav = Navigation(driver)
     nav.set_awesome_bar()
-    sleep(2)
+    WebDriverWait(driver, WAIT_TIMEOUT).until(lambda d: True)
```
philimon-reset, line 35: Would it be possible to implement an explicit wait here with a condition different from the one set? A flat wait for a set amount of time is essentially the same as sleeping.

vsangereanMOZ: I can't get it to work with a different kind of wait since we are just waiting for something that is done by the browser, but we don't see anything change visually. 
Last thing I tried was a def wait_for_awesome_bar_ready(self, timeout=10): kind of solution, but it fails locally.

philimon-reset (approved):
LGTM

PR #574, opened 2025-05-09
Fix linux manual workflow

ben-c-at-moz:
* Run tests in xvfb for Linux manual flow
* Make all fields optional for manual flow (can run each manual flow separately)
* Linux manual flow: Mark and skip tests that fail in xvfb so that we don't get false fails in future runs
* Don't use artifacts on manual runs


whlpatricia (approved):
LGTM!

PR #606, opened 2025-06-10
[Draft]Add Files and selectors for Amazon Germany

vsangereanMOZ:
Does not work yet. 
Need to investigate.


vsangereanMOZ:
Added bug 1971513 for the failed behavior.

PR #611, opened 2025-06-16
vs/Add DE and FR ad and CC files ebay. Small fix for amazon DE

vsangereanMOZ:
Draft.


philimon-reset (changes requested):
The HTML files don't seem to be minified.

soncuteanca:
For me the telephone assertion fails on FR region. I thought initially my condition isdigit() to handle expiration month break these, but I investigated a bit with this conclusions:

PR #645, opened 2025-07-02
Anca/ L10n [CA] - Calvinklein (ad + cc) 

soncuteanca:
### Relevant Links

Bugzilla: [1955328](https://bugzilla.mozilla.org/show_bug.cgi?id=1955328)
TestRail: N/A

- Added [CA] Calvinklein page and locators for address and credit card. The card fields are not working due to browser security policies.
- I've modified a recently added method with a condition to be applied only for expiration month, since in the current state can break other fields.
- "Street address" and "Telephone" fields are not working properly (known bug - [1819901)](https://bugzilla.mozilla.org/show_bug.cgi?id=1819901) and the corresponding tests fail.
- Regarding the existing issues `autofill_name_org `test should not fail (needs more investigation), all the other tests passed locally.

### Process Changes Required

_Mark the relevant boxes:_

- [ ] Adds a dependency (rerun `pipenv install`)
- [ ] Changes the BasePage
- [x] Changes or creates a BOM/POM (name the object model): _
- [ ] Changes CI flow
- [ ] Changes scheduled Beta or DevEdition
- [ ] Changes Git hooks or Github settings
- [ ] Changes L10n harness

### Comments or Future Work

_Do we need to start another PR soon to address something you saw while working on this?_

### Workflow Checklist

- [ ] Please request reviewers
- [ ] If this is an unblocker, please post in Slack.
- [ ] If asked to address comments, please resolve conversations.
- [ ] If asked to change code, please re-request review from the person who wanted changes.

Thank you!


l10n_CM/region/CA.json
```diff
@@ -1,6 +1,12 @@
 {
     "region": "CA",
-    "sites": ["amazon", "walmart", "demo"],
+    "sites": [
+    "amazon",
+    "walmart",
+    "demo",
+    "etsy",
+    "calvinklein"
```
philimon-reset, line 8: I wouldn't add calvinklein here since both the cc and ad tests are skipped. no point in reporting it

modules/page_object_autofill.py
```diff
@@ -245,9 +245,14 @@ def verify_form_data(self, sample_data: CreditCardBase | AutofillAddressBase):
                 )
             elif attr_name == "expiration_date" and len(autofilled_field_value) > 5:
                 autofilled_field_value = autofilled_field_value.replace("20", "")
+
             elif attr_name == "country":
                 expected_value = self.util.get_country_local_translation(expected_value)
-            if autofilled_field_value.isdigit():
+            if (
+                attr_name == "expiration_month"
```
philimon-reset, line 252: i don't think this change is necessary. as long as the field is a digit, the conversion should go through so the other checks don't seem nessecary

philimon-reset (approved):
All good 👍🏽

PR #810, opened 2025-09-24
tracy/ Find "log" instead of "pro"

Tracy-Walker:
#### Relevant Links
Bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1988735
TestRail: https://mozilla.testrail.io/index.php?/tests/view/7428980

#### Description of Code / Doc Changes
Previous fix of this test broke in 144.0b5 due to changes in the about:about page content, though only on Windows.  I suspect Windows profiling work is in flux thus causing instability in this test. This fix adjusts the test to no longer find/match on "pro" and instead key on "log"

#### Workflow Checklist
- [X] Please request reviewers


ben-c-at-moz (approved):
lgtm

PR #1256, opened 2026-04-17
Compare testid properly

arash-e-at-moz:

- Fix comparing testid with the one in manifest

### Workflow Checklist

Thank you!

### DIFF
--- a/scripts/notify_stability_state_changes.py
+++ b/scripts/notify_stability_state_changes.py
@@ -311,7 +311,7 @@ def build_slack_blocks(
                 "type": "section",
                 "text": {
                     "type": "mrkdwn",
-                    "text": f"*{header_text}*\nNo test state changes detected."
+                    "text": f"*{header_text}*\nNo test state changes detected.",
                 },
             }
         ]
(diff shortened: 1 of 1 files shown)


scripts/notify_stability_state_changes.py
```diff
@@ -219,37 +220,48 @@ def build_changes(
...
-        new_state = item["state"]  # Computed state from BQ
+        new_state = item["state"]
         statuses = item["statuses"]
-        current_state = manifest_map.get(nodeid)  # Current state from manifest
+
+        # Try exact match first
+        current_state = manifest_map.get(nodeid)
+
+        # Fall back to file-level path if exact match is missing
+        if current_state is None and "::" in nodeid:
```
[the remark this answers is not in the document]
ben-c-at-moz: I'm not sure this really clarifies anything, but if this is better for you, go for it.

scripts/notify_stability_state_changes.py
```diff
@@ -264,17 +276,36 @@ def build_changes(
...
+    platform: str,
+    runs: int,
+    insufficient_data_count: int,
 ) -> List[dict]:
     """Turn the list of changes into Slack Blocks"""
 
     header_text = f"Stability state changes for {platform} (last {runs} runs)"
 
     if not changes:
+        if insufficient_data_count > 0:
```
ben-c-at-moz, line 293: Do we really need a notification if states _didn't_ change?

ben-c-at-moz (approved):
One question, but the code looks good.

scripts/notify_stability_state_changes.py
arash-e-at-moz: That is a very good question. I assume since we run this workflow manually on demand and usually we end up with at least some changes, it could have minimal impact. Let's see how this turns out to be, and if we end up with this becoming noise in the channel, we can always rectify that.

PR #1318, opened 2026-05-13
Ben/general pyautogui

ben-c-at-moz:
### Relevant Links

Bugzilla: [key](https://bugzilla.mozilla.org/show_bug.cgi?id=2018322), [mouse](https://bugzilla.mozilla.org/show_bug.cgi?id=2018324), [Linux runner](https://bugzilla.mozilla.org/show_bug.cgi?id=1938185)

From @jwmossmoz:
* Create a dummy filepicker to inject for Linux tests
* Update linux runner to 24.04
* Make it possible to install pyautogui / ydotool on Linux
* Update Readme and docs

From @ben-c-at-moz:
* Update filepicker helper to use pyautogui
* Update cv reference images
* Make gui automation calls through a genericized wrapper around pyautogui / whatever else we use in future
* Update tests
* Update key
* Docs and lint
* Remove pynput dependency

### Process Changes Required

- [x] Removes a dependency (rerun `pipenv install`)
- [x] Changes the BasePage
- [x] Changes or creates a BOM/POM (name the object model): PrintPreview, AboutLogins, GenericPdf

In Ubuntu 22.04, we were able to automate the GUI, even on a machine running Wayland natively, presumably because either the compatibility layer (XWayland) worked in a way that it doesn't on the 24.04 images, or because the `-wayland` image, either in the build or in something inherent in 22.04 allowed an X server to start. Without that X server running, neither Pynput nor PyAutoGUI work out of the box. We attempted to use the PyPI `wayland-automation` package, but that also failed. It seems possible that `ydotool` or `dotool` would work, but jmoss proposed the filepicker injection strategy, and that worked better for us.

The downside of this, of course, is that we'll have to inject a modal for every area of test where GUI automation is our prevailing strategy, OR implement `ydotool` directly using that the new `.gui` attribute on the `BasePage`.

### Comments or Future Work

* Investigate whether the modal injection strategy could work for us in other areas of test
* Investigate whether it is worth committing time to PR PyAutoGui with code to automate Wayland

### Workflow Checklist

Thank you!


manifests/key.yaml
```diff
@@ -626,15 +626,21 @@ menus:
     - smoke
     - ci
   test_frequently_used_context_menu:
-    result: pass
+    result:
+      linux: unstable
```
arash-e-at-moz, line 637: Do we need a Bugzilla link?

manifests/key.yaml
```diff
@@ -626,15 +626,21 @@ menus:
...
     splits:
     - functional1
   test_hyperlink_context_menu:
     result: pass
     splits:
     - smoke
   test_image_context_menu_actions:
-    result: pass
+    result:
+      linux: unstable
```
arash-e-at-moz, line 649: Do we need Bugzilla link?

arash-e-at-moz (dismissed):
Some questions left.

taskcluster/kinds/run-smoke-tests/kind.yml
```diff
@@ -27,9 +27,19 @@ tasks:
       command: |-
         export PATH=$HOME/.local/bin:$PWD:$PATH
         mkdir -p artifacts;
-        pip3 install 'pipenv==2023.11.15';
+        pip3 install 'pipenv==2023.11.15' --break-system-packages;
         mv ./config/ci_pyproject.toml ./pyproject.toml;
+        curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
```
arash-e-at-moz, line 32: I believe 1.3 is the latest and greatest TLS version.

tests/drag_and_drop/test_paste_image_text.py
```diff
@@ -91,7 +89,7 @@ def test_paste_image_text(driver: Firefox, sys_platform, temp_selectors):
     web_page.actions.click_and_hold(start_element).move_to_element(
         end_element
     ).release().perform()
-    web_page.copy_selection(keyboard, "paragraph1")
+    web_page.copy_selection("paragraph1")
```
arash-e-at-moz: is "paragraph1" a constant?

taskcluster/kinds/run-smoke-tests/kind.yml
ben-c-at-moz: This I believe is being done on purpose, or rather, it was a change from RelOps

tests/drag_and_drop/test_paste_image_text.py
ben-c-at-moz: I think that's out of scope for this PR. Yes, we should refactor drag_and_drop tests some time.

manifests/key.yaml
ben-c-at-moz: Yep. Sorry.

arash-e-at-moz (approved):
No new comment on my end.

PR #1399, opened 2026-06-04
fix TC kinds

ben-c-at-moz:
### Relevant Links

no link

* Update pipenv install on all kind files for TC

### Process Changes Required

- [x] Changes CI flow
- [x] Changes scheduled Beta / DevEdition / RC

* When we added Pyautogui, we updated our linux runs to execute on Ubuntu 24.04 with Wayland, which required `--break-system-packages` for pipenv (as an all-user install), but we didn't port that code to the schedule runs. This diff fixes that.

### Comments or Future Work

Well, we need to make sure it works on a scheduled run, and we'll want to watch each new type (devedition, rc, etc) of run closely.

### Workflow Checklist

- [x] If this is an unblocker, a message has been posted to #dte-automation in Slack.

Thank you!

### DIFF
--- a/taskcluster/kinds/new-beta-func/kind.yml
+++ b/taskcluster/kinds/new-beta-func/kind.yml
@@ -36,9 +36,13 @@ tasks:
         export STARFOX_SPLIT="${STARFOX_SPLIT}$(($(date +%-I) % $MAX_SPLITS + 1))"
         export PATH=$HOME/.local/bin:$PWD:$PATH
         mkdir -p artifacts;
-        pip3 install 'pipenv==2023.11.15';
+        pip3 install 'pipenv==2023.11.15' --break-system-packages;
         mv ./config/ci_pyproject.toml ./pyproject.toml;
+        curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+        export PATH=$PATH:$HOME/.cargo/bin
+        . "$HOME/.cargo/env"
         pipenv install;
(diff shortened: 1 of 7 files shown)


taskcluster/kinds/new-beta-func/kind.yml
```diff
@@ -36,9 +36,13 @@ tasks:
         export STARFOX_SPLIT="${STARFOX_SPLIT}$(($(date +%-I) % $MAX_SPLITS + 1))"
         export PATH=$HOME/.local/bin:$PWD:$PATH
         mkdir -p artifacts;
-        pip3 install 'pipenv==2023.11.15';
+        pip3 install 'pipenv==2023.11.15' --break-system-packages;
         mv ./config/ci_pyproject.toml ./pyproject.toml;
+        curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
```
arash-e-at-moz: TLS v1.3 is the latest, is there any reason we not using that one?

arash-e-at-moz (dismissed):
One comment/question left.

taskcluster/kinds/new-beta-func/kind.yml
ben-c-at-moz:
```
> curl --proto '=https' --tlsv1.3 -sSf https://sh.rustup.rs
curl: (4) A requested feature, protocol or option was not found built-in in this libcurl due to a build-time decision.
```

arash-e-at-moz (dismissed):
Thanks!

PR #1430, opened 2026-06-18
add test for creating, resizing, moving and deleting drawed areas

vsangereanMOZ:
### Relevant Links

Bugzilla: [2009708](https://bugzilla.mozilla.org/show_bug.cgi?id=2009708)
TestRail: [1938259](https://mozilla.testrail.io/index.php?/cases/view/1938259)

Add coverage for PDF drawing areas by creating a drawing with the Draw toolbar button, selecting the created drawing, deleting it through the PDF.js context menu, and verifying it is removed from the document.
Add GenericPdf helpers for selecting PDF drawing areas and covers deleting a created drawing area through the context menu.

Thank you!

### DIFF
--- a/manifests/key.yaml
+++ b/manifests/key.yaml
@@ -1053,6 +1053,10 @@ pdf_viewer:
       win: pass
     splits:
     - functional2
+  test_pdf_drawing_area_actions:
+    result: pass
+    splits:
+    - functional1
   test_pdf_dropdown_functionality:
     result: pass
     splits:
--- a/modules/page_object_generics.py
+++ b/modules/page_object_generics.py
@@ -314,7 +314,7 @@ class GenericPdf(BasePage):
         self.element_has_text("added-text", text)
         return self
 
-    def get_element_rect(self, element: str | WebElement) -> dict[str, float]:
+    def get_element_rect(self, element: WebElement) -> dict[str, float]:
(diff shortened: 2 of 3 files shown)


ben-c-at-moz (commented):
I do think it's worth looking into the LLM comments, it does look like `select_drawing_area` and `get_drawing_area` return the same element, so you can delete 425 and have 426 be `drawing_area = self.select_drawing_area()`. Also before we merge this I think we need an explanation of `get_drawing_resize_handle()` in the PR conversation section

vsangereanMOZ:
get_drawing_resize_handle() finds the resize handle for the drawing annotation.
It does not resize the drawing by itself. It only returns the small handle element that another method can drag.

Runs JavaScript inside the PDF viewer page. This is used because the resize handle is not easy/reliable to fetch with a normal Selenium selector. The JavaScript inside was created with AI. 
The logic behind it is that the JavaScript receives the drawing area element, calculates its bottom-right corner, then searches the document for visible elements that look like resize handles.
From those candidates, it returns the element closest to the drawing’s bottom-right corner, which should be the handle used for resizing.

ben-c-at-moz (approved):
Thanks for the explanatory note, go ahead and merge if you're comfortable with the bot comments.

PR #1140, opened 2026-03-05
Ai kill switch

jsegunosidefox:
## Delete this block after confirming that you will do the following:

* If asked to make changes, I will re-request reviews.
* If conversations are added, I will resolve them before merging.
* I will delete my branch after merging, or set to delete upon merge.
* I will update any Bugzilla bugs or TestRail test cases if needed.

### Relevant Links

Bugzilla: _
TestRail: _

### Description of Code / Doc Changes

_Leave a bullet-pointed list of changes you made._

### Process Changes Required

_Mark the relevant boxes, delete irrelevant lines._

- [ ] Adds a dependency (rerun `pipenv install`)
- [ ] Modifies a git hook (rerun `./devsetup.sh`)
- [ ] Changes the BasePage
- [x] Changes or creates a BOM/POM (name the object model): _
- [ ] Changes CI flow
- [ ] Changes scheduled Beta or DevEdition
- [ ] Changes Git hooks or Github settings
- [ ] Changes L10n harness

### Screenshots or Explanations

_If you need to explain your code, do it here._

### Comments or Future Work

_Do we need to start another PR soon to address something you saw while working on this?_

### Workflow Checklist

- [ ] Reviewers have been requested.
- [ ] Code has been linted and formatted.
- [ ] If this is an unblocker, a message has been posted to #dte-automation in Slack.

Thank you!

PR #1508, opened 2026-07-15
AI Controls (kill switch) test suite

jsegunosidefox:
### Relevant Links

Bugzilla: _n/a_
TestRail: S71443 (GenAI Settings / AI Controls)

Ports the AI Controls ("kill switch") tests onto `main` (the follow-up to the merged minimal PoC #1302; supersedes the stale #1140). 22 test cases under `tests/ai_controls/`, one file per TestRail case.

- **`conftest.py`** — enterprise-policy support: a `policies_list` fixture writes `distribution/policies.json` next to the binary before launch and removes it in teardown (skips gracefully with a clear message if the dir isn't writable, e.g. a SIP-protected app bundle).
- **`modules/page_object_prefs.py`** — `AboutPrefs` AI helpers: `navigate_to_ai_controls`, `verify_ai_controls_core_elements_visible`, `set_ai_blocking`, `get_ai_killswitch_state`, translations/chatbot getters+setters, `cancel_ai_killswitch_click`, `get_extensions_ml_enabled`. State is read via `aria-pressed` (the inner `input` has no `.pressed` property).
- **`modules/data/about_prefs.components.json`** — added `ai-control-link-preview-select`, `ai-control-smart-tab-groups-select`, `ai-control-pdf-alt-text-select`.
- **`manifests/key.yaml`** — registered `ai_controls` cases.
- Addressed prior review feedback: **C3341331** now asserts `extensions.ml.enabled` flips on block (via the UI toggle — a raw pref set doesn't run the killswitch side effects); **C3340562** exercises the real "Block all AI enhancements?" Cancel path.

Cases that map to inherently manual/gated scenarios (screen readers, non-English/region builds, `about:inference` model storage, high-contrast legibility, telemetry pings, enterprise policy) are left `disabled`/`unstable` rather than given false-confidence assertions.

### Process Changes Required

- [ ] Adds a dependency (rerun `pipenv install`)
- [ ] Changes the BasePage
- [x] Changes or creates a BOM/POM (name the object model): **AboutPrefs** (`modules/page_object_prefs.py`)
- [ ] Changes CI flow
- [ ] Changes scheduled Beta or DevEdition
- [ ] Changes Git hooks or Github settings
- [ ] Changes L10n harness

Verified locally on Firefox Nightly 154: **18 pass / 7 skip / 0 fail** across `tests/ai_controls`. Skips are the 5 policy tests (SIP-protected build dir; run in CI) and 2 non-English `@pytest.mark.skip` cases.

### Workflow Checklist

- [ ] Reviewers have been requested.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jsegunosidefox:
Addressed the automated review (pushed as an amend):

- **`import json`** → moved to module top in `conftest.py`.
- **`set_ai_blocking`** → docstring now states it's a pref-only shortcut that skips UI side effects; points to `toggle_ai_killswitch_click` (used by C3341331).
- **`cancel_ai_killswitch_click`** → matches `label == "Cancel"` (symmetric with the merged `toggle_ai_killswitch_click`'s `== "Block"`) instead of `!= "Block"`.
- **C3276003 / C3276004** → bodies replaced with explicit `pytest.skip("not implemented")` so they can't pass trivially if un-disabled.
- Imports sorted (`ruff --select I`).

Deferred, with reasoning:
- **Redundant `navigate_to_ai_controls()`** — the `about_prefs` fixture pre-opens the pane and the merged C3308998 relies on that contract; the extra navigation is harmless, so I left it rather than risk that test.
- **Dialog button via dedicated JSON key / stable attribute** — kept label matching to stay consistent with the already-merged `toggle_ai_killswitch_click`; worth a follow-up if we want to de-fragilise both together.

Re-verified on Firefox Nightly 154: **16 pass / 9 skip / 0 fail**.

jsegunosidefox:
## Summary (revised)

Per Ben's feedback, this PR now contains **only tests that actually run and assert real behavior** — no stubs, no `pytest.skip`, no `unstable`, no environment-gated cases. The stub/deferred/unautomatable cases from the original port (high-contrast, screen-reader, localization, non-en builds, enterprise policies, sidebar sync, model deletion) have been removed, along with the now-dead harness and selectors they needed.

Result: `conftest.py` and `about_prefs.components.json` are back to **matching `main`** (no shared-infra changes). The PR is just the AI Controls helpers on `AboutPrefs` + the tests that use them + the manifest block.

### Tests — all pass on Firefox Nightly, en-US (8 collected / 8 passed / 0 skipped)
| Case | Asserts |
|---|---|
| c3276002 | core AI Controls elements (toggle, chatbot select, translations select) are present/visible |
| c3298824 | Block-all persists after enabling an individual feature |
| c3310314 | an enabled feature remains enabled after blocking + unblocking via the toggle |
| c3340562 | cancelling the Block-all confirmation leaves AI enabled |
| c3341331 | the WebExtensions ML API is disabled while AI is blocked, restored after unblock |

### Notes for review
- Kill-switch flows go through the **real UI toggle** (`toggle_ai_killswitch_click`) so side effects fire; direct pref writes are used only where the UI control is deliberately hidden, and that trade-off is documented on `set_ai_blocking`.
- Dialog buttons are matched by English label to stay **consistent with the already-merged `toggle_ai_killswitch_click`** (`"Block"`); a locale-robust selector should replace both together in a follow-up rather than diverging here.

_(Superseded my earlier summary above, which described the pre-trim version.)_

tests/ai_controls/test_c3276003_ai_settings_visible_high_contrast.py
ben-c-at-moz, line 1: Just a heads up, we don't need stubs for tests we've determined were unsuitable for automation, so it would be best to remove this.

tests/ai_controls/test_c3276002_ai_features_keyboard_accessible.py
ben-c-at-moz, line 1: The linked test indicates that each element should be tested with the keyboard, which we can do, but is not present in this test, just FYI

conftest.py
```diff
@@ -501,6 +514,33 @@ def driver(
     options.binary_location = fx_executable
     # options.set_preference("app.update.disabledForTesting", False)
 
+    # Enterprise policies live in distribution/policies.json next to the binary.
```
ben-c-at-moz: Interesting and potentially very useful. Have we confirmed that automated sessions respect the policy files if they have been written like this?

ben-c-at-moz (commented):
I see you haven't requested reviewers, which may mean you're not ready with this PR, but I have some comments you should take a look at before you get to that point.

modules/page_object_prefs.py
```diff
@@ -1344,6 +1344,181 @@ def expect_ai_selects_state(self, disabled=False) -> BasePage:
...
+        side effects (e.g. flipping extensions.ml.enabled). Use
+        toggle_ai_killswitch_click when those side effects must be exercised
+        (see C3341331).
+
+        Arguments:
+            block: True to block AI enhancements, False to allow them.
+        """
+        if block != self.get_ai_killswitch_state():
+            self.driver.execute_script(
+                "Services.prefs.setStringPref('browser.ai.control.default', arguments[0]);",
```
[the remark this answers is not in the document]
ben-c-at-moz: Worth looking at.

modules/page_object_prefs.py
```diff
@@ -1344,6 +1344,181 @@ def expect_ai_selects_state(self, disabled=False) -> BasePage:
...
+        """
+        Click the AI killswitch toggle and dismiss the confirmation dialog with
+        its "Cancel" button, leaving AI unblocked. Mirrors
+        toggle_ai_killswitch_click but exercises the cancel path of the
+        "Block all AI enhancements?" prompt.
+        """
+        self.click_on("ai-controls-toggle")
+        self.element_visible("ai-controls-disable-dialog-button")
+        buttons = self.get_elements("ai-controls-disable-dialog-button")
+        cancel = [el for el in buttons if el.get_attribute("label") == "Cancel"]
```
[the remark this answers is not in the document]
ben-c-at-moz: Also worth a look.

tests/ai_controls/test_c3277712_disable_all_policies.py
ben-c-at-moz, line 1: Looks good, let's double check that this is doing exactly what we expect, but Policy-level stuff is definitely demoable.

tests/ai_controls/test_c3277713_disable_chatbot_policies.py
```diff
@@ -0,0 +1,38 @@
...
+    C3277713 - Disable only chatbot via policies
+    Requires enterprise policy to disable only the chatbot feature.
+    """
+    about_prefs.navigate_to_ai_controls(verify=False)
+
+    # Chatbot control should be hidden when the chatbot policy is active
+    about_prefs.element_not_visible("ai-control-sidebar-chatbot-select")
+
+    # Other features should remain available
+    state = about_prefs.get_ai_translations_state()
```
ben-c-at-moz: At some point we might want to make helper functions that `expect` a given state, so we're not relying on `assert`, which has the tendency to be flaky.

tests/ai_controls/test_c3341230_pdf_alt_text_non_en.py
ben-c-at-moz, line 1: sadly, this may be out of scope for STARfox at the moment, as we don't run any regular tests against non-en builds, and it's not immediately on the roadmap

tests/ai_controls/test_c3341234_link_preview_localized.py
ben-c-at-moz, line 1: We may be able to get access to the locale data, and change it using the pref setter fixtures

tests/ai_controls/test_c3341325_provider_selection_reflects.py
ben-c-at-moz, line 1: Is it not possible for us to do this?

conftest.py
```diff
@@ -501,6 +514,33 @@ def driver(
...
+        try:
+            policy_file.unlink()
+        except OSError as exc:
+            logging.warning(
+                f"Could not remove stale policies file {policy_file}: {exc}"
+            )
+    if policies_list:
+        try:
+            dist_dir.mkdir(parents=True, exist_ok=True)
+            policy_file.write_text(json.dumps({"policies": policies_list}, indent=2))
```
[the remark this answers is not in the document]
ben-c-at-moz: Worth looking into

tests/ai_controls/test_c3276002_ai_features_keyboard_accessible.py
ben-c-at-moz: To do: update the test so that each element is tested with the keyboard.

tests/ai_controls/test_c3298824_block_all_persists.py
ben-c-at-moz, line 1: Good choice, nice and quick test.

tests/ai_controls/test_c3341331_extensions_api_disabled.py
```diff
@@ -0,0 +1,35 @@
...
+
+def test_web_extensions_ai_api_disabled_when_blocking(about_prefs: AboutPrefs):
+    """
+    C3341331 - Web extensions AI API disabled when blocking
+    """
+    about_prefs.navigate_to_ai_controls()
+
+    # Precondition: the WebExtensions ML API is enabled before we block, so the
+    # transition below is meaningful (fails loudly here if the default changes).
+    about_prefs.expect(lambda _: about_prefs.get_extensions_ml_enabled() is True)
```
ben-c-at-moz: minor, but when we're checking against a value other than `None`, it's better to use `==` in comparisons.

modules/page_object_prefs.py
```diff
@@ -1362,6 +1344,121 @@ def expect_ai_selects_state(self, disabled=False) -> BasePage:
...
+
+    def set_ai_translations(self, state: str) -> BasePage:
+        """
+        Set the AI Translations feature state.
+
+        Arguments:
+            state: "available" or "blocked"
+        """
+        select_elem = self.get_element("ai-control-translations-select")
+        self.driver.execute_script(
```
ben-c-at-moz, line 1631: Is there a reason we need to do JS-injection here? I'm ok if there is, but if there's not, I'd prefer we do it with regular Selenium / BasePage

modules/page_object_prefs.py
```diff
@@ -1362,6 +1344,121 @@ def expect_ai_selects_state(self, disabled=False) -> BasePage:
...
+
+    def get_ai_translations_state(self) -> str:
+        """
+        Get the current state of the AI Translations feature.
+
+        Returns:
+            str: Current state ("available" or "blocked")
+        """
+        select_elem = self.get_element("ai-control-translations-select")
+        return self.driver.execute_script("return arguments[0].value;", select_elem)
```
ben-c-at-moz: I would prefer to use a Selenium method rather than JS injection here.

ben-c-at-moz (changes requested):
Just a few things to look at, but I think it's pretty close. To get the tests to pass again, merge a fresh copy of main.

tests/ai_controls/test_c3276002_ai_features_keyboard_accessible.py
```diff
@@ -0,0 +1,93 @@
...
+
+from modules.page_object_prefs import AboutPrefs
+
+
+@pytest.fixture()
+def test_case():
+    return "3276002"
+
+
+def _focused_matches(driver, element) -> bool:
```
ben-c-at-moz, line 21: Fine for now, but if we need this again, we'll want to factor it out into the prefs page object model or the BasePage

ben-c-at-moz (dismissed):
This is fine to merge once the CI passes. You may need to rebase to current `main` for that to work

ben-c-at-moz (approved):
alright, let's merge this!

PR #1514, opened 2026-07-20
update conftest to run tests in UI

ben-c-at-moz:
Emergency PR - Geckodriver 0.37.1 breaks our main Firefox session setup. This PR fixes that only.


arash-e-at-moz (dismissed):
Good to go!

ben-c-at-moz:
Because MacOS uses Selenium's gecko picker (and my attempts to override it seem to have failed, we'll be running on 0.37.0 in Mac until the next cycle). This is fine.

PR #1530, opened 2026-07-23
Ben/harness fixes

ben-c-at-moz:
### Relevant Links

un-ticketed QoL improvement

* Attempt to fix TC CI runs targeting the wrong diff
* Reduce conditions for l10n tests firing in CI
* Reduce test set that runs on main conftest etc.

### Process Changes Required

- [x] Changes CI flow

Currently, TC seems to get a different value for `git diff` than GHA, resulting in different test sets run in Linux. Also, changes to conftest cause the smoke suite to run, which is a lot of tests.

### Workflow Checklist

Thank you!

### DIFF
--- a/manifests/key.yaml
+++ b/manifests/key.yaml
@@ -11,6 +11,7 @@ address_bar_and_search:
     result: pass
     splits:
     - smoke
+    - ci-extended
   test_added_open_search_engine_default:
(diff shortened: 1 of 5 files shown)


Tracy-Walker (approved):
lgtm

PR #1554, opened 2026-08-05
vs/Stabilize reopen tabs through keys.

vsangereanMOZ:
### Relevant Links

Bugzilla: N/A
TestRail: N/A

Navigated the original tab to about:mozilla so Firefox does not reuse a blank tab during restoration.
Tracked newly opened and restored tabs using their exact window handles.
Closed and reopened tabs individually, waiting for the expected tab count after each action.
Verified that every shortcut restores exactly one new tab.
Corrected the keyboard shortcut helper to release Shift and the platform modifier properly.
Validated that the four restored tabs contain exactly the expected URLs.

Thank you!

### DIFF
--- a/modules/browser_object_tabbar.py
+++ b/modules/browser_object_tabbar.py
@@ -445,12 +445,10 @@ class TabBar(BasePage):
         return selected_tabs
 
     @BasePage.context_chrome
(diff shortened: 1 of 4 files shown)


sv-hyacoub (approved):
Looks good.

PR #1559, opened 2026-08-12
Prep STARfox to handle nightly runs

ben-c-at-moz:
### Relevant Links

JIRA: [ticket](https://mozilla-hub.atlassian.net/browse/DTE-4367)

In preparation to handle Nightly as the system under test, we need to make a few changes to the codebase:
* Create a pytest fixture that returns "Firefox" or "Nightly" based on the name of the application
* Create a `nightly` split for functional testing
* Temporary workflow changes while prepping for the full Nightly workflow:
  * allow split name to be passed into main.yml workflow dispatch
  * Attempt to allow test exclusion based on overlapping splits
* Update tests that failed because of reliance on "Firefox" app name

### Process Changes Required

- [x] Changes CI flow

### Comments or Future Work

Headed testing in GHA linux doesn't work, even with the test exclusions. I'm going ahead and calling that "out of scope" for this change, but we're going to have to revisit it.

### Workflow Checklist

- [x] Code has been linted and formatted..

Thank you!

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -517,7 +517,7 @@ jobs:
           "$FX_EXECUTABLE" --version;
           XVFB_PID=$(ps -e | grep vfb | head -1 | awk '{print $1}')
           if [[ -n "$XVFB_PID" ]]; then kill -9 "$XVFB_PID"; fi;
-          xvfb-run -e /dev/stdout -f "$HOME/.Xauthority" -s "-screen 0 1600x1200x24" uv run pytest --fx-executable="$FX_EXECUTABLE" --geckodriver=geckodriver $(cat selected_tests) || TEST_EXIT_CODE=$?;
+          xvfb-run -f "$HOME/.Xauthority" -s "-screen 0 1600x1200x24" uv run pytest --fx-executable="$FX_EXECUTABLE" --geckodriver=geckodriver $(cat selected_tests) || TEST_EXIT_CODE=$?;
           exit $TEST_EXIT_CODE
 
       - name: Run Smoke Tests in Ubuntu (Headed)
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -506,6 +506,8 @@ jobs:
         id: select
         run: |
           uv run python -m scripts.choose_test_split
+        env:
+          STARFOX_EXCLUDE: no_xvfb
 
       - name: Run Smoke Tests in Ubuntu
         if: ${{ steps.select.conclusion == 'success' && inputs.dry_run != true }}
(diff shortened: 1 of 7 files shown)


tests/tabs/test_display_customize_button.py
```diff
@@ -35,7 +35,7 @@ def test_customize_button_displayed_in_tab_bar(driver: Firefox):
     with driver.context(driver.CONTEXT_CHROME):
         customize_tab = tabs.get_tab(2)
         assert customize_tab is not None, "Customize tab should still exist."
-        assert tabs.get_tab_title(customize_tab) == "Customize Firefox"
+        assert tabs.get_tab_title(customize_tab) == f"Customize {browser_name}"
```
arash-e-at-moz, line 38: I wonder what is this change for?

conftest.py
```diff
@@ -571,6 +571,14 @@ def driver(
             driver.quit()
 
 
+@pytest.fixture()
+def browser_name(driver: Firefox) -> str:
+    """
+    Is it 'Firefox' or 'Nightly'?
```
arash-e-at-moz, line 577: I am not sure if Nightly is comparable to Firefox for browser_name? Any thoughts?

scripts/choose_test_split.py
```diff
@@ -164,6 +164,12 @@ def dedupe(run_list: list) -> list:
         run_list = dedupe(run_list)
         run_list = manifest.filter_filenames_by_pass(run_list)
 
+        if os.environ.get("STARFOX_EXCLUDE"):
+            print(f"Excluding tests from split: {os.environ['STARFOX_EXCLUDE']}...")
+            for exclude in dedupe(manifest.gather_split(os.environ["STARFOX_EXCLUDE"])):
```
arash-e-at-moz, line 170: What is the `STARFOX_EXCLUDE` logic for?

conftest.py
ben-c-at-moz: Yep, the issue is that in some tests, we're expecting "Firefox is doing X" or "Click to update Firefox", but in nightly it says "Nightly is doing X"

scripts/choose_test_split.py
ben-c-at-moz: Excluding tests we don't want to run. For example, you may have an environment where certain tests are likely to fail, but you don't want to mark them unstable, because they work in all other environments. In our case, this really just means "works in Linux TC, fails in Linux GHA"

tests/tabs/test_display_customize_button.py
ben-c-at-moz: see below

arash-e-at-moz (dismissed):
Good to go!

.github/workflows/main.yml
```diff
@@ -84,8 +89,10 @@ jobs:
         shell: pwsh
         run: |
           "MANUAL=true" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
-          "STARFOX_SPLIT=smoke" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
+          "STARFOX_SPLIT=$env:TEST_SET" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
```
arash-e-at-moz, line 92: Seems confusing as we are using word SPLIT and TEST_SET interchangeably. Might be better to consistently use only one of them?

arash-e-at-moz (commented):
Only one tiny change, otherwise LGTM.

.github/workflows/main.yml
ben-c-at-moz: I'll fix that, re-review request incoming

arash-e-at-moz (approved):

PR #1574, opened 2026-08-20
Akash/ Stabilize test_download_pdf_data

akashshawdev:
### Relevant Links

Bugzilla: [2064111](https://bugzilla.mozilla.org/show_bug.cgi?id=2064111)
TestRail: [1020327](https://mozilla.testrail.io/index.php?/cases/view/1020327)

* Mock the file picker on all platforms instead of only Linux, so the native "Save As" panel never opens.
* Added `dom.disable_beforeunload` and removed the `hard_quit` fixture, so the browser shuts down cleanly.
* Re-enabled `test_download_pdf_data` in `key.yaml`.

### Process Changes Required

- [ ] Changes the BasePage
- [ ] Changes CI flow

The test drove a native "Save As" panel with pyautogui. The image match against
`data/cv/mac_save_button.png` never succeeded, so every run fell through to a blind `Enter`
that landed on whichever window had focus — hence the flakiness. `install_mock_file_picker`
swaps `nsIFilePicker` in-process, so no OS dialog appears. TestRail step 4 expects that dialog
on screen; it no longer renders, but `wait_for_mock_file_picker` asserts Firefox invoked the
picker, which the old code never checked.

The edited PDF also raised a save prompt on shutdown, blocking `driver.quit()` until
geckodriver's 60s timeout and orphaning a Firefox process each run. `dom.disable_beforeunload`
suppresses it, making `hard_quit` unnecessary. Runtime: 64s → 4s.

Tested headed on macOS and Windows 11, Nightly 156 and Release 154 on both.
PR CI covers Windows, macOS, and Linux.

### Comments or Future Work

* No test now covers the real OS save dialog. One dedicated test driving it via accessibility
  APIs (`AXUIElement` / UIAutomation) would beat image matching — mac and Windows only, since
  Wayland blocks synthetic input. Happy to file if reviewers agree.
* `test_download_pdf_with_form_fields` (bug 2064142) and `test_image_context_menu_actions` still
  call `handle_os_download_confirmation`. Once both move to the mock, that helper and
  `data/cv/*_save_button.png` can go.
* `downloads_folder` (`conftest.py:244`) uses `%USERNAME%` while `home_folder` uses
  `%HOMEDRIVE%%HOMEPATH%`; they disagree when the account name and profile directory
  differ, which broke local Windows runs. CI is unaffected. Filing separately.
* Step 3 of the case also lists checkboxes and dropdowns; this test only fills a text field.

### Workflow Checklist

Thank you!

### DIFF
--- a/manifests/key.yaml
+++ b/manifests/key.yaml
@@ -1134,8 +1134,7 @@ pdf_viewer:
     splits:
     - functional2
   test_download_pdf_data:
-    comment: https://bugzilla.mozilla.org/show_bug.cgi?id=2064111
-    result: unstable
+    result: pass
     splits:
     - functional2
   test_download_pdf_with_form_fields:
--- a/tests/pdf_viewer/test_download_pdf_data.py
+++ b/tests/pdf_viewer/test_download_pdf_data.py
@@ -78,8 +78,3 @@ def test_download_pdf_data(
     tabs.open_and_switch_to_new_tab()
     driver.get("file://" + os.path.realpath(saved_pdf_location))
     pdf_viewer.element_visible("edited-name-field")
-
-    # Verify if the file exists
-    assert os.path.exists(saved_pdf_location), (
-        f"The file was not downloaded to {saved_pdf_location}."
-    )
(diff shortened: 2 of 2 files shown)


ben-c-at-moz (approved):
This is fine -- if we're mocking the filepicker in all OSes, though, I wonder if we even need this test. It should be covered by a test in the main tree.

PR #1576, opened 2026-08-20
Hani/ Remove unstable from password manager us test

sv-hyacoub:
### Relevant Links

Bugzilla: [2064167](https://bugzilla.mozilla.org/show_bug.cgi?id=2064167)

* Remove unstable password_manager/test_password_manager_on_google_US on mac since it passed every time I ran the test locally.

### Process Changes Required

- [ ] Changes the BasePage
- [ ] Changes CI flow

N/A

### Comments or Future Work

N/A

### Workflow Checklist

Thank you!

### DIFF
--- a/manifests/key.yaml
+++ b/manifests/key.yaml
@@ -1056,11 +1056,7 @@ password_manager:
     - functional2
     - nightly
   test_password_manager_on_google_US:
-    comment: https://bugzilla.mozilla.org/show_bug.cgi?id=2064167, old mac bug
-    result:
-      linux: pass
-      mac: unstable
-      win: pass
+    result: pass
     splits:
     - functional2
   test_password_manager_on_reddit:


soncuteanca:
I don't think we should just flip the flag without any work done, the test is flaky somewhere since it failed in beta 1.
Also, the CI on this PR didn't actually run the test. The diff is key.yaml only, and choose_test_split.py ignores manifest changes, so it fell back to the ci split. And even if it had run and come back green, one pass wouldn't really rule out some underlying issue.
It's just an opinion, could well be a low rate of flakiness, and sometimes the right fix is hard to pin down. I'll leave this comment and wait for a second opinion.

Tracy-Walker:
Anca, your opinion is correct.  This one is sort of a one off. Ben filed this bug against the Mac problem because the key.yaml file only had a comment about the test being unstable on Mac GHA.  Without a bug on file, the issue was never going to be investigated.  

Hani, trigger this test in CI by making some comment change in the test case and push here. If it passes, I'll approve. If not, you'll have some logs/info to investigate the test failure.

ben-c-at-moz (approved):

PR #1627, opened 2026-09-04
Smart Window (S70279) test suite

jsegunosidefox:
### Relevant Links

Bugzilla: N/A
TestRail: S70279 (AI Smart Window) — C3248785, C3248786, C3309849, C3309851, C3309853, C3309854, C3310319

- Adds `tests/smart_window/` — 7 tests, the first slice of Smart Window automation
  - `c3248785` — switching to Smart while signed out routes to the FxA flow (`entrypoint=smartwindow`)
  - `c3248786` — a Smart Window can be switched back to Classic, and Smart-only chrome is torn down
  - `c3309849` — switching windows preserves the active tab and its page
  - `c3309851` — Private Browsing offers no Smart Window entry point
  - `c3309853` — New Private Window from a Smart Window is private, not Smart
  - `c3309854` — New Classic Window from a Smart Window is Classic
  - `c3310319` — blocking Smart Window in AI Controls removes the Switch Windows button
- Adds `modules/browser_object_smart_window.py` + `modules/data/smart_window.components.json`
- Adds `set_ai_smart_window` / `get_ai_smart_window_state` to `AboutPrefs`, plus the
  `ai-control-smart-window-select` entry
- Adds hamburger entries for New Smart Window / New Classic Window to `panel_ui.components.json`
- Registers the suite in `manifests/key.yaml` (`functional1`)

### Process Changes Required

- [x] Changes or creates a BOM/POM (name the object model): **`SmartWindow`** (new), `AboutPrefs` and `PanelUi` (additive only)

**Smart Window does not need a signed-in FxA account to reach.** Setting
`browser.smartwindow.enabled` makes the feature available, and `AIWindow.toggleAIWindow` puts a
window into the Smart state — the same call the product makes *after* it authorizes the user. The
auth gate lives only in `AIWindowAccountAuth.ensureAIWindowAccess`, reached via `launchWindow`.

`SmartWindow.activate_smart_window` wraps that call and documents when **not** to use it: where the
entry point itself is the subject, assert the sign-in redirect instead. `c3248785` does exactly
that — it goes through the real button and asserts the FxA handoff.

**Coverage caveat worth a reviewer's attention:** three tests exercise fully real user paths; four
use the activation fixture, so they verify behaviour inside and out of a Smart Window but **not the
authenticated route into one**. Nothing here covers the real sign-in → Smart Window flow; that
still needs a signed-in profile or manual testing.

**No stubs.** Verified with a negative control — with `browser.smartwindow.enabled` set to `false`,
all 7 fail. That check caught a real bug: `switcher_button_available` was reporting a hidden button
as present. Disabling the pref leaves the element in the DOM with `hidden=true`, while blocking via
AI Controls destroys the widget outright, so the predicate has to test visibility, not existence.

Verified on Custom Firefox **156.0b3** (the build CI uses) and **155.0.1**, 3 consecutive runs plus
combined runs with `tests/ai_controls/` (14/14).

### Comments or Future Work

- **`C3248785` asserts the signed-out gate**, not the switch as TestRail describes it (that case
  assumes a signed-in profile). The case needs annotating or splitting.
- **`C3309858` "Open a link from Smart to Classic Window" does not match the build.** In 155.0.1 and
  156.0b3 the link context menu offers only "Open Link in New Smart Window"
  (`context-openlinksmartwindow`); there is no Classic counterpart. Needs a product answer.
- **Next highest-leverage work is a smart bar input harness**, which unblocks ~12 cases
  (C3248361/2, C3321908, C3322034-36, C3322041, C3977124/6, C3322037/40/42). The smart bar input is
  a ProseMirror contenteditable two shadow roots deep, and Marionette's `shadow_root` API throws on
  it, so `send_keys` never lands.
- Restore Session (C3309859-61, C4065796) is reachable next — SessionStore persists `isAIWindow` and
  restore is gated on `isAIWindowEnabled()`, not re-auth.
- `C3917535` (enterprise policy) needs a `policies.json` harness built from scratch; the kill-switch
  one was removed in #1508.

### Workflow Checklist

### DIFF
--- a/modules/browser_object_smart_window.py
+++ b/modules/browser_object_smart_window.py
@@ -1,0 +1,212 @@
+import logging
+
+from selenium.webdriver.common.keys import Keys
+
+from modules.page_base import BasePage
+
+AI_WINDOW_MODULE = "moz-src:///browser/components/aiwindow/ui/modules/AIWindow.sys.mjs"
+
+
+class SmartWindow(BasePage):
+    """
+    Browser Object Model for the Smart Window (AI Window) chrome.
+
+    Covers the Switch Windows button and its Classic/Smart panel, plus the
+    chrome that only exists while a window is in the Smart Window state.
+    """
+
+    URL_TEMPLATE = "about:blank"
+
+    # ── Smart Window state ───────────────────────────────────────────────
+
+    @BasePage.context_chrome
+    def is_smart_window_active(self) -> bool:
+        """
+        Report whether the current window is in the Smart Window state.
+
+        Reads AIWindow.isAIWindowActive rather than the `ai-window` attribute
+        so the check follows the product's own definition of "active".
+        """
+        return self.driver.execute_script(
+            f"""
+            const {{ AIWindow }} = ChromeUtils.importESModule("{AI_WINDOW_MODULE}");
+            return AIWindow.isAIWindowActive(window);
+            """
+        )
+
+    def expect_smart_window_active(self, active: bool = True) -> BasePage:
+        """
+        Wait until the current window is (or is not) a Smart Window.
+        """
+        self.expect(lambda _: self.is_smart_window_active() == active)
+        return self
+
+    @BasePage.context_chrome
+    def activate_smart_window(self) -> BasePage:
+        """
+        Put the current window into the Smart Window state.
+
+        TEST HARNESS ONLY. This calls AIWindow.toggleAIWindow directly, which
+        is the same call the product makes *after* it has authorized the user.
+        Going through the UI instead (the Switch Windows button, the hamburger
+        menu, the AI settings link) routes through
+        AIWindowAccountAuth.ensureAIWindowAccess, which needs a live FxA
+        account and so cannot run in this suite.
+
+        Use this to reach Smart Window surfaces whose *behaviour* is under
+        test. Do not use it in a test whose subject is the entry point itself
+        -- assert the sign-in redirect instead (see C3248785).
+        """
+        logging.info("Activating Smart Window state via AIWindow.toggleAIWindow")
+        self.driver.execute_script(
+            f"""
+            const {{ AIWindow }} = ChromeUtils.importESModule("{AI_WINDOW_MODULE}");
+            AIWindow.toggleAIWindow(window, true, "other");
+            """
+        )
+        self.expect_smart_window_active(True)
+        return self
+
+    # ── Switch Windows button ────────────────────────────────────────────
+
+    def open_window_switcher(self) -> BasePage:
+        """
+        Click the Switch Windows button and wait for its panel to open.
+        """
+        self.click_on("window-switcher-button")
+        self.element_visible("window-switcher-view")
+        return self
+
+    @BasePage.context_chrome
+    def close_window_switcher(self) -> BasePage:
+        """
+        Dismiss the Switch Windows panel with ESC.
+        """
+        self.actions.send_keys(Keys.ESCAPE).perform()
+        self.element_not_visible("window-switcher-view")
+        return self
+
+    def get_switcher_selection(self) -> str:
+        """
+        Return which entry the Switch Windows panel shows as current.
+
+        Returns
+        -------
+        str
+            "smart", "classic", or "none" if neither is checked.
+        """
+        for name, label in (
+            ("switch-to-smart", "smart"),
+            ("switch-to-classic", "classic"),
+        ):
+            # The panel sets state with toggleAttribute, so an unselected entry
+            # has no attribute at all. Reject "false" anyway, so a future
+            # checked="false" would not read as selected.
+            checked = self.get_element(name).get_attribute("checked")
+            if checked and checked != "false":
+                return label
+        return "none"
+
+    def expect_switcher_selection(self, expected: str) -> BasePage:
+        """
+        Wait until the Switch Windows panel marks `expected` as current.
+        """
+        self.expect(lambda _: self.get_switcher_selection() == expected)
+        return self
+
+    def switch_to_classic_window(self) -> BasePage:
+        """
+        Switch the current window to a Classic Window via the Switch Windows
+        button, and wait for the switch to take effect.
+        """
+        self.open_window_switcher()
+        self.click_on("switch-to-classic")
+        self.expect_smart_window_active(False)
+        return self
+
+    def click_switch_to_smart_window(self) -> BasePage:
+        """
+        Click the Smart entry in the Switch Windows panel.
+
+        Does not wait for a Smart Window: while signed out this navigates to
+        the FxA sign-in page instead.
+        """
+        self.open_window_switcher()
+        self.click_on("switch-to-smart")
+        return self
+
+    @BasePage.context_chrome
+    def switcher_button_available(self) -> bool:
+        """
+        Report whether the Switch Windows button is offered in this window.
+
+        Checks visibility, not merely existence. The three ways Smart Window
+        can be unavailable do not look alike in the DOM:
+
+        - Private Browsing window -> the widget is never built (no element)
+        - browser.smartwindow.enabled=false -> element exists with hidden=true
+        - blocked in AI Controls -> the widget is destroyed (no element)
+
+        An existence-only check would call the middle case "available", which
+        would let a test pass with the feature switched off.
+
+        Drops the implicit wait so an absent button returns immediately rather
+        than costing a full timeout.
+        """
+        original = self.driver.timeouts.implicit_wait
+        self.driver.implicitly_wait(0)
+        try:
+            elements = self.get_elements("window-switcher-button")
+            return bool(elements) and elements[0].is_displayed()
+        finally:
+            self.driver.implicitly_wait(original)
+
+    # ── Tabs ─────────────────────────────────────────────────────────────
+
+    @BasePage.context_chrome
+    def get_selected_tab_url(self) -> str:
+        """
+        Return the URL of the selected tab, read from chrome.
+
+        Used instead of driver.current_url where the action under test opens a
+        new tab, since the content-context handle can still point at the old
+        one.
+        """
+        return self.driver.execute_script(
+            "return gBrowser.selectedBrowser.currentURI.spec;"
+        )
+
+    def expect_selected_tab_url_contains(self, fragment: str) -> BasePage:
+        """
+        Wait until the selected tab's URL contains `fragment`.
+        """
+        self.expect(lambda _: fragment in self.get_selected_tab_url())
+        return self
+
+    # ── Windows ──────────────────────────────────────────────────────────
+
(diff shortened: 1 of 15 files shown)


tests/smart_window/test_c3248785_switch_to_smart_window_prompts_signin.py
```diff
@@ -0,0 +1,37 @@
...
+    smart_window.open_window_switcher()
+    smart_window.expect_switcher_selection("classic")
+    smart_window.close_window_switcher()
+
+    smart_window.click_switch_to_smart_window()
+
+    # Sign-in is required first: the FxA flow opens in a new tab, tagged with
+    # the Smart Window entrypoint.
+    smart_window.expect_selected_tab_url_contains(FXA_SIGN_IN_HOST)
+    signin_url = smart_window.get_selected_tab_url()
```
ben-c-at-moz, line 32: Not sure why we're not using `expect_selected_tab_url_contains()` instead of the below assert statements

modules/browser_object_smart_window.py
```diff
@@ -0,0 +1,212 @@
...
+    def expect_smart_window_active(self, active: bool = True) -> BasePage:
+        """
+        Wait until the current window is (or is not) a Smart Window.
+        """
+        self.expect(lambda _: self.is_smart_window_active() == active)
+        return self
+
+    @BasePage.context_chrome
+    def activate_smart_window(self) -> BasePage:
+        """
```
ben-c-at-moz, line 46: I would clarify this comment--I believe it can be pared down to a couple of lines.

modules/browser_object_smart_window.py
```diff
@@ -0,0 +1,212 @@
...
+        Does not wait for a Smart Window: while signed out this navigates to
+        the FxA sign-in page instead.
+        """
+        self.open_window_switcher()
+        self.click_on("switch-to-smart")
+        return self
+
+    @BasePage.context_chrome
+    def switcher_button_available(self) -> bool:
+        """
```
ben-c-at-moz, line 140: Simplify comment

modules/browser_object_smart_window.py
```diff
@@ -0,0 +1,212 @@
...
+    def expect_selected_tab_url_contains(self, fragment: str) -> BasePage:
+        """
+        Wait until the selected tab's URL contains `fragment`.
+        """
+        self.expect(lambda _: fragment in self.get_selected_tab_url())
+        return self
+
+    # ── Windows ──────────────────────────────────────────────────────────
+
+    def wait_for_new_window(self, known_handles: set[str]) -> str:
```
ben-c-at-moz, line 188: Do we need this, when we have `wait_for_num_windows()` in BasePage?

tests/smart_window/test_c3309849_switch_windows_while_visiting_website.py
```diff
@@ -0,0 +1,41 @@
...
+    C3309849 - Switch windows while visiting a website
+    """
+    tabs = TabBar(driver)
+    driver.get(URL)
+    tabs.expect_title_contains("Gort")
+
+    smart_window.activate_smart_window()
+
+    # The page the user was on survives the switch to Smart.
+    assert driver.current_url == URL
```
ben-c-at-moz, line 34: use `smart_window.url_contains(URL)` to avoid flakiness.

tests/smart_window/test_c3309849_switch_windows_while_visiting_website.py
```diff
@@ -0,0 +1,41 @@
...
+    tabs.expect_title_contains("Gort")
+
+    smart_window.switch_to_classic_window()
+
+    # ...and the switch back to Classic.
+    assert driver.current_url == URL
```
ben-c-at-moz, line 40: see above

tests/smart_window/test_c3309851_smart_window_missing_in_private_browsing.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def test_smart_window_missing_in_private_browsing(
+    driver: Firefox, smart_window: SmartWindow
+):
+    """
+    C3309851 - Smart Window option missing in Private Browsing
+    """
+    # Baseline: the normal window offers the Smart option.
+    assert smart_window.switcher_button_available()
```
ben-c-at-moz, line 26: I think we should wait for this.

tests/smart_window/test_c3309851_smart_window_missing_in_private_browsing.py
```diff
@@ -0,0 +1,41 @@
...
+    existing_handles = set(driver.window_handles)
+    panel_ui.open_private_window()
+    driver.switch_to.window(smart_window.wait_for_new_window(existing_handles))
+
+    private_window = SmartWindow(driver)
+    private_window.is_private()  # waits and asserts; raises on timeout
+
+    # The Switch Windows button is not built at all in a private window, so
+    # there is no entry point to a Smart Window.
+    assert not private_window.switcher_button_available()
```
ben-c-at-moz, line 41: We should wait for this too. `element_exists()` and `element_does_not_exist()`

tests/smart_window/test_c3309853_open_private_window_from_smart_window.py
```diff
@@ -0,0 +1,48 @@
...
+
+    new_handle = active_smart_window.wait_for_new_window(existing_handles)
+    driver.switch_to.window(new_handle)
+
+    private_window = SmartWindow(driver)
+    private_window.is_private()  # waits and asserts; raises on timeout
+
+    # Private Browsing never carries the Smart Window state.
+    private_window.expect_smart_window_active(False)
+    assert not private_window.switcher_button_available()
```
ben-c-at-moz, line 41: wait instead of assert

tests/smart_window/test_c3309853_open_private_window_from_smart_window.py
```diff
@@ -0,0 +1,48 @@
...
+
+    # Positive control: the window we opened it from is still a Smart Window,
+    # so the assertions above reflect Private Browsing being exempt rather
+    # than Smart Window being unavailable in this profile.
+    driver.switch_to.window(smart_handle)
+    active_smart_window.expect_smart_window_active(True)
+    assert active_smart_window.switcher_button_available()
```
ben-c-at-moz, line 48: wait instead of assert, `element_exists` and `element_does_not_exist`

tests/smart_window/test_c3310319_block_smart_window_from_ai_controls.py
```diff
@@ -0,0 +1,35 @@
...
+def test_block_smart_window_from_ai_controls(driver: Firefox):
+    """
+    C3310319 - Enable/Block Smart Window from AI controls
+    """
+    smart_window = SmartWindow(driver)
+    about_prefs = AboutPrefs(driver, category="ai")
+    about_prefs.navigate_to_ai_controls()
+
+    # The feature starts available, so the window switcher is offered.
+    assert about_prefs.get_ai_smart_window_state() == "available"
```
ben-c-at-moz, line 28: you can use wait with a lambda here, or change the function to wait for the window state instead of just getting it.

tests/smart_window/test_c3310319_block_smart_window_from_ai_controls.py
```diff
@@ -0,0 +1,35 @@
...
+    assert smart_window.switcher_button_available()
```
ben-c-at-moz, line 29: wait instead of assert

ben-c-at-moz (changes requested):
A few things here, mostly comment length, some re-development of functions, and a little overuse of `assert` instead of a wait function. But all of the tests look good, and the structure of the BOM is good.

jsegunosidefox:
Re-request: addressed the review feedback in 2af55a5b.

- Test-level `assert`s replaced with existing waits (`url_contains`, `expect_selected_tab_url_contains`, `expect_smart_window_active`, `element_visible` / `element_does_not_exist`).
- Dropped `wait_for_new_window` — call sites now use `wait_for_num_windows` + `switch_to_new_window`.
- Dropped `switcher_button_available` — tests use `element_visible` / `element_does_not_exist` directly.
- Added `AboutPrefs.expect_ai_smart_window_state`, matching the `expect_ai_killswitch_state` pattern.
- Pared down the `activate_smart_window` docstring.

Net: +41 / -96 across 8 files, no new features. Pre-flight review clean.

modules/browser_object_smart_window.py
```diff
@@ -0,0 +1,151 @@
...
+
+    def get_switcher_selection(self) -> str:
+        """
+        Return which entry the Switch Windows panel shows as current.
+
+        Returns
+        -------
+        str
+            "smart", "classic", or "none" if neither is checked.
+        """
```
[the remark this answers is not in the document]
ben-c-at-moz: This is correct and I recommend `@context_of_model` in the really unlikely case that this gets moved to like a prefs page or something.

tests/smart_window/conftest.py
```diff
@@ -0,0 +1,37 @@
+import pytest
+
+from modules.browser_object_smart_window import SmartWindow
```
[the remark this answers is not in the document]
ben-c-at-moz: model output is correct here

tests/smart_window/test_c3310319_block_smart_window_from_ai_controls.py
```diff
@@ -0,0 +1,35 @@
...
+@pytest.fixture()
+def test_case():
+    return "3310319"
+
+
+def test_block_smart_window_from_ai_controls(driver: Firefox):
+    """
+    C3310319 - Enable/Block Smart Window from AI controls
+    """
+    smart_window = SmartWindow(driver)
```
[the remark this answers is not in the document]
ben-c-at-moz: model output is correct

ben-c-at-moz (changes requested):
The fixes are minor, but it would be good to get the tests to pass.

jsegunosidefox:
Round 2 addressed in 1ae6a9a5:

- Added `@BasePage.context_of_model` to `get_switcher_selection` so it's self-contained rather than relying on outer decorators.
- Switched all `smart_window/` imports to the aggregate `modules.browser_object`, per CLAUDE.md.
- `test_c3310319` now uses the `smart_window` fixture from conftest instead of constructing `SmartWindow(driver)` inline.

Suite verified 7/7 passing locally after these changes.

jsegunosidefox:
Round 3 addressed (16cab7e5):

- `activate_smart_window` no longer decorates with `@BasePage.context_chrome` — chrome is scoped to just the `execute_script` block, so `expect_smart_window_active`'s poll loop no longer triple-nests chrome context.
- The `AboutPrefs` aggregate-import fix from the same review round already landed as cedf6826.

Suite verified 7/7 passing locally.

ben-c-at-moz (approved):
Alright, good to merge

PR #1641, opened 2026-09-09
Akash/ Fix image context menu URL check and save flake

akashshawdev:
### Relevant Links

Bugzilla: [2070538](https://bugzilla.mozilla.org/show_bug.cgi?id=2070538)
TestRail: --

* Updated the expected image URL pattern. Wikipedia moved thumbnails to `thumb.wikimedia.org` and added `?utm_*` params, breaking the hardcoded host on every platform.
* `test_save_image_as` now mocks the file picker on all platforms, not just Linux. Driving the real OS save dialog was the cause of the Mac and Windows instability.
* Dropped the `@pytest.mark.headed` marker, since the test no longer needs desktop input.
* * Flipped `manifests/key.yaml` from `unstable` on Linux and Windows to `result: pass`, since the native dialog that caused the instability is gone.

### Process Changes Required
--

The URL check uses `re.match`, which only anchors the start, so the trailing query string does not need to be matched. Only the host needed to become flexible: `[\w-]+\.wikimedia\.org`.

All 3 subtests pass locally on Mac and Windows, headed and headless.

### Comments or Future Work

None

### Workflow Checklist

### DIFF
--- a/tests/menus/test_image_context_menu_actions.py
+++ b/tests/menus/test_image_context_menu_actions.py
@@ -25,7 +25,7 @@ LINK_IMAGE_URL = (
 )
 # The thumbnail host varies (upload/thumb) and may have a query string.
 LOADED_IMAGE_URL = (
-    r"https://[\w-]+\.wikimedia\.org/wikipedia/commons/thumb/a/a0/Firefox_logo%2C_2019\.svg/\d+px"
+    r"https://[a-zA-Z0-9-]+\.wikimedia\.org/wikipedia/commons/thumb/a/a0/Firefox_logo%2C_2019\.svg/\d+px"
     r"-Firefox_logo%2C_2019\.svg\.png"
 )
 # Match the saved file by stem; the served format varies (.webp/.png).
--- a/tests/menus/test_image_context_menu_actions.py
+++ b/tests/menus/test_image_context_menu_actions.py
@@ -1,6 +1,5 @@
 import os
 import re
-from time import monotonic, sleep
 
 import pytest
 from selenium.webdriver import Firefox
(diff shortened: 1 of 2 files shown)


tests/menus/test_image_context_menu_actions.py
```diff
@@ -64,13 +62,11 @@ def test_open_image_in_new_tab(driver: Firefox):
     wiki_image_page.verify_opened_image_url("wikimedia", LOADED_IMAGE_URL)
 
 
-@pytest.mark.headed
-def test_save_image_as(driver: Firefox, sys_platform, downloads_folder, delete_files):
+def test_save_image_as(driver: Firefox, downloads_folder, delete_files):
```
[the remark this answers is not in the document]
akashshawdev: Done

tests/menus/test_image_context_menu_actions.py
```diff
@@ -24,18 +23,17 @@ def delete_files_regex_string():
 LINK_IMAGE_URL = (
     "https://en.wikipedia.org/wiki/Firefox#/media/File:Firefox_logo,_2019.svg"
 )
+# The thumbnail host varies (upload/thumb) and may have a query string.
 LOADED_IMAGE_URL = (
-    r"https://upload\.wikimedia\.org/wikipedia/commons/thumb/a/a0/Firefox_logo%2C_2019\.svg/\d+px"
+    r"https://[\w-]+\.wikimedia\.org/wikipedia/commons/thumb/a/a0/Firefox_logo%2C_2019\.svg/\d+px"
```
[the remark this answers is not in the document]
akashshawdev: Applied

tests/menus/test_image_context_menu_actions.py
arash-e-at-moz: Better to capture the newly created tab handle instead of assuming it is window_handles[1].

tests/menus/test_image_context_menu_actions.py
```diff
@@ -80,53 +76,22 @@ def test_save_image_as(driver: Firefox, sys_platform, downloads_folder, delete_f
...
-    # Linux CI can't reliably drive the native "Save As" dialog, so use the mock
-    # file picker, which is the repo convention for downloads. Other platforms
-    # confirm the native dialog.
-    use_mock_picker = sys_platform == "Linux"
-    mock_saved_image_location = os.path.join(downloads_folder, SAVED_IMAGE_FILENAME)
-    if use_mock_picker:
-        wiki_image_page.install_mock_file_picker(mock_saved_image_location)
-
+    # mock the native save dialog, which CI cannot drive reliably
+    saved_image_location = os.path.join(downloads_folder, SAVED_IMAGE_FILENAME)
```
arash-e-at-moz: Please use `pathlib.Path` rather than `os.path`.

arash-e-at-moz (changes requested):

tests/menus/test_image_context_menu_actions.py
akashshawdev: Done, now diffing handles before/after

tests/menus/test_image_context_menu_actions.py
akashshawdev: Done, switched to pathlib.Path

arash-e-at-moz (approved):
Good to go.

PR #1656, opened 2026-09-15
Anca/ Glean stabilization - Ecosia CI 

soncuteanca:
### Relevant Links

Bugzilla: [2071318](https://bugzilla.mozilla.org/show_bug.cgi?id=2071318)
TestRail: N/A

- Follow up to https://github.com/mozilla/fx-desktop-qa-automation/pull/1650:
    - There have been some changes in how CI handles Ecosia. The Ecosia failures were Cloudflare bot challenges, not a Firefox regression, the same tests on different engines pass. Two gaps let them slip past the existing detector:
    - The challenge was gone by the time we looked. Abandonment flows close the tab or navigate away, so by the time a failure was classified the interstitial was no longer on screen. We now record it while the SERP is still up and fall back to that record.
    - The Cloudflare token only matched in the URL. `__cf_chl_tk` leaves the URL after the redirect but stays in the page body as a referer, so the URL-only marker caught just the tests that failed mid-redirect. Matching the body catches the rest.
    - The Ecosia cases stay as they are for now and I will watch them a while before deciding. The challenge looks like an IP cooldown: the first run after a pause passed all of them, and the bot only returns once they are being run regularly. Depending on whether they turn out to be Blocked non-stop or only intermittently, we will either mark them unsuitable or would potentially give the IP time to go clean. also, note that the shorter release cycle leaves the manual team less time to run these regularly, so randomizing them looks like the better option., but will decide later.
### Process Changes Required

- [ ] Changes the BasePage
- [ ] Changes CI flow

### Comments or Future Work

_Do we need to start another PR soon to address something you saw while working on this?_

### Workflow Checklist

Thank you!


ben-c-at-moz (approved):
LGTM, but don't merge unless you've confirmed that a manual test of the functionality succeeds.

soncuteanca:
- I've checked them manually, and also Cristi Baica from search team confirmed nothing changed on the manual front.
- It's Cloudflare, 17 of our 24 Ecosia cases pass and record their metrics correctly; the 7  that fail all hit a Cloudflare challenge, the flows are shared and the same ones pass across other engines.
- The challenge comes in two forms with the same page title. One is non-interactive and clears itself in about two seconds; the other is a "Verify you are human" checkbox that never clears. Which one you get depends on IP reputation, and hammering it escalates  from the first to the second.
- Letting the check complete on a throwaway navigation before the measured search fixed  4 of the cases cleanly. It can't do anything about the interactive form, and using a throwaway navigation brings no guarantees on parallel runs.
- It's not CI-specific either, I get challenged from my own machine too after a while during investigation.
- The problem is that, letting them run and marked them Blocked if an external challenge is triggered is not ideal on the long term. Retrying doesn't help; it just keeps the IP warm and makes the escalation more likely. Maybe put them on hold for two weeks and retest them on a calm "reputation" .. not sure is enough though since there are plenty of Ecosia runs in the meantime. Maybe worth a try. Note that since the release cycle is so short manual team won't have enough bandwidth to test them often either, so a randomization could be a solution.

PR #1663, opened 2026-09-16
vs/Add test for PDF text location.

vsangereanMOZ:
### Relevant Links

Bugzilla: [2009712](https://bugzilla.mozilla.org/show_bug.cgi?id=2009712)
TestRail: [1938265](https://mozilla.testrail.io/index.php?/cases/view/1938265)

Simple test add. Just a few steps, as in the testrail.

Thank you!


modules/page_object_generics.py
```diff
@@ -465,15 +465,63 @@ def draw_on_pdf_page(
...
+            const left = Math.max(0, Math.min(rect.left, rect.right));
+            const right = Math.min(window.innerWidth, Math.max(rect.left, rect.right));
+            const top = Math.max(0, Math.min(rect.top, rect.bottom));
+            const bottom = Math.min(window.innerHeight, Math.max(rect.top, rect.bottom));
+            return {
+                x: Math.floor((left + right) / 2) + arguments[1],
+                y: Math.floor((top + bottom) / 2) + arguments[2],
+            };
+            """,
+            page,
```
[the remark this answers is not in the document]
vsangereanMOZ: only used for the generic PDF

modules/page_object_generics.py
```diff
@@ -467,15 +465,63 @@ def draw_on_pdf_page(
...
+    def expect_text_at_pdf_page_location(
+        self,
+        page_number: str,
+        x_offset: int,
+        y_offset: int,
+        tolerance: int = 5,
+    ) -> BasePage:
+        """Verify the text editor contains the point selected on the PDF page."""
+        page = self.get_element("pdf-page", labels=[page_number])
+        selected_point = self.driver.execute_script(
```
soncuteanca: I think the bot is right: steps 1 and 2 each measure "where's the middle of the page" separately, at two different moments in time and this can drift. Maybe, instead of recalculating Point B from scratch, just remember Point A from step 1 and reuse that exact same number in step 2. That way you're always comparing against the real click, not a fresh guess.

soncuteanca: As per DTE Code Contribution Expectations,  a comment explaining the function of a js script is always required, if its use can't be avoided.

vsangereanMOZ: Bot mentioned that these steps are used in another test, not this. 
For your comment I will save "last_pdf_text_point" in the init of Generic PDF.

vsangereanMOZ: js execution here is the only solution. will add short explanation.

modules/page_object_generics.py
```diff
@@ -465,15 +466,64 @@ def draw_on_pdf_page(
...
         return self
 
+    def expect_text_at_pdf_page_location(
+        self,
+        tolerance: int = 5,
+    ) -> BasePage:
+        """Verify the text editor contains the point selected on the PDF page."""
+        assert self._last_pdf_text_point is not None, (
+            "Expected text to be added before checking its location."
+        )
```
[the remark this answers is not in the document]
vsangereanMOZ: we don't use -o. 
Not in project scope.

soncuteanca (approved):

PR #1682, opened 2026-09-18
Smart Bar harness + C3321887 (Go button dropdown opens)

jsegunosidefox:
### Relevant Links

Bugzilla: N/A
TestRail: S70279 (AI Smart Window) — C3321887

- Adds `modules/browser_object_smart_bar.py` (`SmartBar`) + `modules/data/smart_bar.components.json`
- Exports `SmartBar` from the aggregate `modules/browser_object.py`
- Adds `tests/smart_window/test_c3321887_go_button_dropdown_opens.py`
- Registers it in `manifests/key.yaml` (`functional1`)

### Process Changes Required

- [x] Changes or creates a BOM/POM (name the object model): **`SmartBar`** (new), `browser_object.py` (one export line)

**Why this BOM scripts its traversal instead of using selectors.** The Smart Bar renders inside
`<browser id="ai-window-browser">`, whose document is `chrome://browser/content/aiwindow/aiWindow.html`,
with the editor two shadow roots down (`ai-window` → `moz-multiline-editor`) and the Go button's
`input-cta` on a separate branch of the same document. Neither a `<browser>` contentDocument nor those
shadow chains are reachable through the `components.json` selector system, so the accessors script the
traversal. Each script is a single traversal or a single property read/write on the product's own
element APIs — commented per the DTE guidance on `execute_script`.

**Two things worth a reviewer's eye in the traversal.** The `cta()` walk starts from the browser's
*document*, not from `ai-window`'s shadow root — `input-cta` is not inside the latter, and searching
there finds nothing. And the match uses `el.matches("input-cta")` rather than a `tagName` comparison,
because these are HTML-namespaced elements in a XUL document, so `tagName` reads back as
`"html:input-cta"`.

**What C3321887 asserts.** That opening the Go split button's dropdown reaches the *open state*, then
that the Search With submenu lists more than one engine. It deliberately does not assert specific menu
item presence: the items are what the `open_action_menu` wait keys on, so asserting them afterwards
could never fail. The `len(engines) > 1` check is locale-agnostic, and its failure message calls out
that an empty list means the submenu lazily renders.

Verified on Custom Firefox **157.0b3**. C3321887 **5/5** in isolation and **3/3** on the final commit;
full `tests/smart_window/` suite **8/8**.

### Comments or Future Work

- **C3322034** ("delete inline tag with backspace") was written against this harness and then **pulled
  from this PR**: it ran 2/5 on repeat runs. Shipping it marked `unstable` was not the right call, so
  it and the @-mention accessors it needed are held back until the flakiness is understood.
- **C3322036** ("maximum number of added tags") does not reproduce — tagged 6 tabs with no cap
  enforced and no "Max Tags reached" message. Needs a product answer before it can be written.
- **C3322035 / C3322041** are blocked on the context chip above the text field
  (`website-chip-container`) never being populated in 157.0b3 regardless of tagging; its `websites`
  property stays empty. Raised as a product question rather than asserted here.
- Five further Smart Bar cases (C3248361/62, C3321908, C3977124/26) are unblocked by this harness and
  are the natural follow-up PR.

### Workflow Checklist


modules/browser_object_smart_bar.py
```diff
@@ -0,0 +1,214 @@
+import logging
+
+from modules.page_base import BasePage
+
+# The Smart Bar renders inside <browser id="ai-window-browser">, whose document
```
ben-c-at-moz, line 5: This comment could be streamlined.

modules/browser_object_smart_bar.py
```diff
@@ -0,0 +1,214 @@
...
+# is chrome://browser/content/aiwindow/aiWindow.html, and the editor sits two
+# shadow roots down (ai-window -> moz-multiline-editor). Neither a <browser>
+# contentDocument nor that shadow chain is reachable through the components.json
+# selector system, so the accessors below script the traversal instead. Each
+# script is a single traversal or a single property read/write on the product's
+# own MultilineEditor API.
+_TRAVERSE = """
+function aiDoc() {
```
ben-c-at-moz, line 13: are we actually sure we need this instead of using `BasePage().switch_to_iframe()`?

modules/browser_object_smart_bar.py
```diff
@@ -0,0 +1,214 @@
...
+
+        Assigns MultilineEditor.value, the element's own API, which replaces
+        the full contents and keeps the ProseMirror document valid. Passing ""
+        clears the field.
+        """
+        logging.info("Setting Smart Bar text to %r", text)
+        # Guarded like get_smart_bar_text: an unguarded assignment throws an
+        # opaque "Cannot set properties of null" if this is ever called
+        # without open_smart_bar() having run first.
+        assigned = self._script(
```
ben-c-at-moz, line 133: Are you sure we can't accomplish this without JS injection?

ben-c-at-moz (changes requested):
A few things to work on, mostly around whether JS injection is actually needed, but the direction is good.

modules/browser_object_smart_bar.py
jsegunosidefox: Streamlined in b6cdd428 — cut from 11 lines to the two facts that matter (where the editor lives, how deep), and folded in the answers to your other two questions so the next reader does not have to ask them again.

modules/browser_object_smart_bar.py
jsegunosidefox: Good question — I tested it rather than assuming, and no, it does not work:

```
driver.switch_to.frame(<ai-window-browser element>)
  -> NoSuchFrameException: Unable to locate frame for element: [object XULFrameElement]
```

`ai-window-browser` is a XUL `<browser>`, not an `<iframe>`, so `switch_to.frame()` has nothing to switch into. I also tried the shadow route:

```
ai_window_element.shadow_root
  -> WebDriverException: Only supported in content context
```

and this UI only exists in chrome, so that API is unavailable too.

Worth flagging one thing I found while checking: the `components.json` `shadowParent` path is not JS-free either. `get_shadow_content` falls back to `execute_script("return arguments[0].shadowRoot.children", element)` (`modules/util.py:834`), and in chrome context `find_shadow_chrome_element` matches by string-searching `outerHTML` on every node (`modules/util.py:877`), restricted to id/tag/class selectors. So the supported path is JS underneath as well — just less direct.

All of this is now recorded in the header comment.

modules/browser_object_smart_bar.py
jsegunosidefox: Partly yes, and I have changed it — thanks for pushing on this.

Typing can be native, and now is. `set_smart_bar_text` focuses the editor and then sends real key events via `perform_key_combo` + `ActionChains`, instead of assigning `MultilineEditor.value`. Verified 5/5. That is a genuine improvement beyond style: assigning `value` bypasses the editor input handling, which is exactly why the @-mention popup renders "No results found" when driven that way — native keys exercise the real path.

The focus call is the one bit I could not remove. Elements inside the `ai-window-browser` contentDocument have no usable Selenium reference from chrome:

```
editor_element.send_keys("text")
  -> StaleElementReferenceException: The element ... is not known in the current browsing context
```

so there is nothing to click or `send_keys` against until something focuses it, and that has to come from privileged JS.

Select-all uses `perform_key_combo(Keys.CONTROL, "a")`, which maps CONTROL to COMMAND on macOS (`page_base.py:473`), so it stays correct on the Linux CI workers.

I also dropped `clear_smart_bar` and `get_placeholder_hints` in the same commit — nothing called either, same reason the @-mention accessors came out earlier.

ben-c-at-moz: alright, that's a good change

modules/data/smart_bar.components.json
```diff
@@ -0,0 +1,22 @@
...
+  "ai-window-browser": {
+    "selectorData": "ai-window-browser",
+    "strategy": "id",
+    "groups": [],
+    "comment": {
+      "description": "Chrome browser element hosting aiWindow.html, created when the Ask button is clicked. The Smart Bar lives inside its contentDocument, behind two shadow roots, so its internals are reached via the SmartBar BOM's scripted accessors rather than selectors here.",
+      "location": "Smart Window chrome"
+    }
+  },
+  "smart-window-ask-button": {
```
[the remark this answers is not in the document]
ben-c-at-moz: Correct, unless it appears in two different places, which it might!

modules/browser_object_smart_bar.py
ben-c-at-moz: Alright, sounds good.

modules/browser_object_smart_bar.py
```diff
@@ -0,0 +1,245 @@
...
+  const d = aiDoc();
+  if (!d) return null;
+  let hit = null;
+  // aiWindow.html nests about four shadow roots deep to reach the CTA; 12 is
+  // headroom against further nesting while still bounding a bad walk.
+  (function walk(node, depth) {
+    if (!node || depth > 12 || hit) return;
+    for (const el of node.querySelectorAll("*")) {
+      // Match with CSS, not tagName: these are HTML-namespaced elements in a
+      // XUL document, so tagName reads back as "html:input-cta".
```
ben-c-at-moz, line 43: true, but we could query selectors on tagname of "html:input-cta"...that's a nitpick, and won't block, but it could end up faster? Maybe?

modules/browser_object_smart_bar.py
```diff
@@ -0,0 +1,245 @@
...
+        Return the Smart Bar's current text, or "" before it is available.
+
+        Returning a value rather than throwing lets expect() poll this during
+        the window where the editor has not finished loading.
+        """
+        return self._script("const e = editor(); return e ? e.value : '';")
+
+    def set_smart_bar_text(self, text: str) -> BasePage:
+        """
+        Replace the Smart Bar's text by typing it.
```
ben-c-at-moz, line 130: I think this block could be a bit more concise

modules/browser_object_smart_bar.py
```diff
@@ -0,0 +1,245 @@
...
+            # correct on the Linux CI workers too.
+            self.perform_key_combo(Keys.CONTROL, "a")
+            self.actions.send_keys(text).perform()
+
+        self.expect_smart_bar_text(text)
+        return self
+
+    def expect_smart_bar_text(self, text: str) -> BasePage:
+        """
+        Wait until the Smart Bar's text equals `text`.
```
ben-c-at-moz, line 159: Likewise, I think this block could be more concise

modules/browser_object_smart_bar.py
```diff
@@ -0,0 +1,245 @@
...
+        focusing it from privileged JS.
+        """
+        logging.info("Setting Smart Bar text to %r", text)
+        focused = self._script(
+            "const p = prosemirror();if (!p) return false;p.focus();return true;"
+        )
+        if not focused:
+            raise AssertionError("Smart Bar editor is not available to write to")
+
+        with self.driver.context(self.driver.CONTEXT_CHROME):
```
ben-c-at-moz, line 148: There's a `BasePage().perform_key_combo_chrome()` so you don't need to explicitly switch contexts.

modules/browser_object_smart_bar.py
```diff
@@ -0,0 +1,245 @@
...
+        self.expect_action_menu_open(True)
+        return self
+
+    def get_action_menu_items(self) -> list[str]:
+        """Return the action menu's panel-item data-l10n-ids, in order."""
+        return self._script("return menuItemIds(ctaLists()[0]);")
+
+    def get_search_with_items(self) -> list[str]:
+        """
+        Return the Search With submenu's entries as visible text, in order.
```
ben-c-at-moz, line 210: This block could be more concise

modules/browser_object_smart_bar.py
```diff
@@ -0,0 +1,245 @@
...
+        otherwise return another menu's contents as though they were engines.
+        """
+        count = self._script("return ctaListCount();")
+        if count != 2:
+            raise AssertionError(
+                f"expected 2 panel-lists in the CTA (actions, Search With), "
+                f"found {count} -- the positional lookup in ctaLists() is no "
+                f"longer safe and needs revisiting"
+            )
+        return self._script(
```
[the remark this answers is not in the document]
ben-c-at-moz: I'm not bothered by this, but the review net output does seem a little more complete? No big deal either way

modules/browser_object_smart_bar.py
```diff
@@ -0,0 +1,213 @@
...
+
+    def get_search_with_items(self) -> list[str]:
+        """
+        Return the Search With submenu's entries as visible text, in order.
+
+        Engine names are not localised strings with l10n ids, so these are
+        read as text.
+        """
+        return self._script(
+            "const l = ctaLists()[1];"
```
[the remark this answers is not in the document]
ben-c-at-moz: This is probably correct

tests/smart_window/test_c3321887_go_button_dropdown_opens.py
```diff
@@ -0,0 +1,56 @@
...
+"""
+
+import pytest
+
+from modules.browser_object import SmartBar, SmartWindow
+from modules.browser_object_smart_bar import (
+    ACTION_MENU_GO_TO_SITE,
+    ACTION_MENU_SEARCH_WITH,
+    ACTION_MENU_SEARCH_WITH_DEFAULT,
+)
```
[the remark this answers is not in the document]
ben-c-at-moz: lol, the net output isn't really different. I recommend:
```python
from modules.browser_object import SmartBar, SmartWindow
...
for expected in (
    SmartWindow.ACTION_MENU_GO_TO_SITE,
...
```

tests/smart_window/test_c3321887_go_button_dropdown_opens.py
```diff
@@ -0,0 +1,60 @@
...
+
+    items = bar.get_action_menu_items()
+    for expected in (
+        ACTION_MENU_GO_TO_SITE,
+        ACTION_MENU_SEARCH_WITH_DEFAULT,
+        ACTION_MENU_SEARCH_WITH,
+    ):
+        assert expected in items, f"{expected} missing from action menu: {items}"
+
+    # Step 2: the Search With… submenu lists the available engines. Its
```
ben-c-at-moz, line 48: This block could be more concise

ben-c-at-moz (changes requested):
We're basically there, just a couple more quick changes.

modules/data/smart_bar.components.json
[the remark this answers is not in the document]
jsegunosidefox: Checked — it is one element, so a single owner is safe here. Both declarations used `"strategy": "id"` against `smartwindow-ask-button`, and an id is unique per document, so they could not have been two different buttons.

`smartwindow-ask-button` now appears exactly once in the repo, in `smart_window.components.json:41`, and `SmartBar.open_smart_bar()` delegates the click to `SmartWindow`.

If the button ever does show up in a second location it will need a distinct id anyway, and that gets its own entry — so this does not paint us into a corner.

tests/smart_window/test_c3321887_go_button_dropdown_opens.py
```diff
@@ -0,0 +1,60 @@
...
+
+    bar.open_action_menu()  # waits for the open state internally
+
+    items = bar.get_action_menu_items()
+    for expected in (
+        ACTION_MENU_GO_TO_SITE,
+        ACTION_MENU_SEARCH_WITH_DEFAULT,
+        ACTION_MENU_SEARCH_WITH,
+    ):
+        assert expected in items, f"{expected} missing from action menu: {items}"
```
ben-c-at-moz, line 46: Are we at all worried this menu won't be populated by the time we get here?

tests/smart_window/test_c3321887_go_button_dropdown_opens.py
```diff
@@ -0,0 +1,60 @@
...
+
+    # Step 2: the Search With… submenu lists the available engines. Its
+    # panel-items are rendered eagerly rather than on expand -- verified by
+    # reading them without expanding, over five consecutive runs. Driving the
+    # submenu open via its own toggle() was tried and is unreliable (1/5).
+    # The case expects "all the available search engines" -- assert several
+    # are listed rather than naming one, since both the default engine and the
+    # offered list vary by locale and region.
+    engines = bar.get_search_with_items()
+    assert len(engines) > 1, (
```
ben-c-at-moz, line 56: Similarly, are we worried we'll reach this line with engines temporarily == 0?

ben-c-at-moz (approved):
A couple of non-blocking questions, merge if you're satisfied the `assert`s won't be flaky

ben-c-at-moz (approved):
LGTM

PR #1683, opened 2026-09-18
Smart Window: AI chat sidebar open/close (C3311329, C3978359)

jsegunosidefox:
### Relevant Links

Bugzilla: N/A
TestRail: S70279 (AI Smart Window) — C3311329, C3978359

- Adds AI chat sidebar accessors to the existing `SmartWindow` BOM
- Adds `tests/smart_window/test_c3311329_toggle_ai_chat_from_ask_button.py`
- Adds `tests/smart_window/test_c3978359_close_sidebar_from_x_button.py`
- Registers both in `manifests/key.yaml` (`functional1`)

### Process Changes Required

- [x] Changes or creates a BOM/POM (name the object model): **`SmartWindow`** (additive only)

**Why `ai_sidebar_open()` measures geometry rather than the `hidden` attribute.** `#ai-window-box`
is present in the DOM before the sidebar has ever been opened, just collapsed, and it collapses
again on close — in both states `hidden` is null. An attribute check would report the sidebar as
open from the moment the window becomes Smart. Measuring the bounding rect is the only signal that
actually tracks the visible state. Verified: `{browser_vis: false, box_vis: false}` before the
first Ask click, `true/true` after, `false/false` after closing.

**Why `close_ai_sidebar()` uses scripted traversal.** The X button lives inside
`<browser id="ai-window-browser">`, several shadow roots below `chrome://browser/content/aiwindow/aiWindow.html`.
Neither a `<browser>` contentDocument nor that shadow chain is reachable via the `components.json`
selector system. It matches on `data-l10n-id="aiwindow-close-sidebar"` rather than the visible
"Close" text so the lookup does not depend on the UI locale, and raises rather than silently
passing if the button is not found.

**Note on independence:** this PR touches only `SmartWindow`, which is already on main. It has no
dependency on #1682 (Smart Bar harness) and the two can merge in either order. The `key.yaml`
inserts are adjacent, so whichever merges second needs a trivial rebase.

Verified on Custom Firefox **157.0b3**. Full `tests/smart_window/` suite **9/9 passing**.

### Comments or Future Work

- During investigation the Ask-button toggle initially appeared not to close the sidebar. That was
  a timing artifact from clicking twice in quick succession — with a settle between clicks it
  closes reliably. Flagging in case the same shape shows up elsewhere.

### Workflow Checklist


modules/browser_object_smart_window.py
```diff
@@ -152,3 +154,154 @@ def expect_selected_tab_url_contains(self, fragment: str) -> BasePage:
...
+
+    def toggle_ai_sidebar(self) -> BasePage:
+        """
+        Click the Ask button, which opens the sidebar when closed and closes
+        it when open.
+        """
+        self.click_on("smart-window-ask-button")
+        return self
+
+    def close_ai_sidebar(self) -> BasePage:
```
ben-c-at-moz, line 185: The comment blocks in this function can be made more concise

modules/browser_object_smart_window.py
```diff
@@ -152,3 +154,154 @@ def expect_selected_tab_url_contains(self, fragment: str) -> BasePage:
...
+            if walk_was_truncated and not clicked_at_least_once:
+                cause += (
+                    ", and the shadow walk hit its depth cap of 12 at least "
+                    "once, so the button may be nested below it"
+                )
+            raise AssertionError(f"AI chat sidebar did not close: {cause}") from None
+        return self
+
+    @BasePage.context_chrome
+    def _click_sidebar_close_button(self) -> Literal[True, False, "truncated"]:
```
ben-c-at-moz, line 240: Comments in this function are also too verbose

modules/browser_object_smart_window.py
```diff
@@ -152,3 +152,93 @@ def expect_selected_tab_url_contains(self, fragment: str) -> BasePage:
...
+                        if (el.shadowRoot) {
+                            walk(el.shadowRoot, depth + 1);
+                            if (hit) return;
+                        }
+                    }
+                })(doc, 0);
+            if (!hit) return false;
+            hit.click();
+            return true;
+        """)
```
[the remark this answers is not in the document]
ben-c-at-moz: this is correct

ben-c-at-moz (changes requested):
A few things, all of them pretty small/quick

ben-c-at-moz (approved):
LGTM!

PR #1692, opened 2026-09-21
Smart Window: reach History from the hamburger menu (C3313457)

jsegunosidefox:
### Relevant Links

Bugzilla: N/A
TestRail: S70279 (AI Smart Window) — C3313457

- Adds `tests/smart_window/test_c3313457_history_from_hamburger_menu.py`
- Adds one element to `modules/data/panel_ui.components.json`: `panel-ui-chats-history`
  (`appMenu-chats-history-button`)
- Registers the test in `manifests/key.yaml` (`functional1`)

### Process Changes Required

None. No BOM/POM class changes — the existing `PanelUi` object already covers the History path;
this only adds one element definition to its manifest.

**Why Chats is asserted before the History subview is opened.** `appMenu-chats-history-button` is
a sibling of `appMenu-history-button` in the hamburger's *main* view, not a child of the History
subview. Once the subview is open the main view is hidden, so asserting it afterwards fails. It is
included because it is the Smart Window's own addition to this menu — without it this would be a
generic Firefox history test that passes in a Classic Window too.

**Why History is clicked directly instead of via `PanelUi.open_history_menu()`.** That helper calls
`open_panel_menu()` itself, and `open_panel_menu()` clicks the button unconditionally
(`browser_object_panel_ui.py:44-52`) — so calling it a second time toggles the panel shut. Clicking
`panel-ui-history` directly after a single `open_panel_menu()` is the same shape
`reopen_recently_closed_tabs` already uses.

**The manifest change is purely additive** — one new key, 9 insertions and 0 deletions. No existing
element definition is touched, so the other suites that consume `panel_ui.components.json` cannot
be affected by it.

Verified on Custom Firefox **157.0b3**. C3313457 **4/4** consecutive; full `tests/smart_window/`
suite **8/8**.

One suite run showed `test_c3309853` failing with the empty-message `TimeoutException` that the
staged-update dialog produces; it passed **3/3** in isolation immediately after and is unrelated to
this change.

### Comments or Future Work

This is the first of the S70279 History group. The rest of that group is **not** in this PR, with
reasons:

- **C3313619 / C3313620** ("from Title Menu") drive the native macOS menu bar, which this repo
  already refuses to test on Darwin (see `tests/session_restore/test_closed_tabs_from_multiple_windows_shown_in_history_menu_bar.py`,
  `skipif(platform.system() == "Darwin")`). I cannot validate them locally, so I would rather not
  ship them unrun.
- **C3313455 / C3313456** ("from Home page") — a Smart Window's home page is
  `chrome://browser/content/aiwindow/aiWindow.html` (`AIWindow.newTabURL`). It does contain the
  entry points (`aiwindow-fullpage-chat-history`, and a `fullpage-chats-menu` panel with
  "View all chats"), but every one of them reports zero width while signed out — the same auth wall
  already recorded for the AI-response cases.
- **C3313459** ("History when tabs are opened") still needs a decision on whether `about:` pages are
  expected to be recorded in places history; deferred rather than guessed at.

### Workflow Checklist


ben-c-at-moz (approved):
LGTM

